Issue #82, opened 2017-02-28
Getting a 2.0 out


kal:
I'd like to get folks views on the next dotNetRDF release both in terms of its scope and its numbering.

Rob had started work on a complete redesign of the DNR API, slating it for a 2.0 release. However not a lot has been done on that recently, although I really do want to get around to it as there is a lot of good stuff in there and scope for really tidying up some of the rougher edges in the 1.0 API.

In the meantime, .NET has moved forwards and we really need to update the packaging of the library to reflect this. That means in particular migrating to .NET Core (see #76), and as a result possibly dropping support both for .NET 3.5 and for signed assemblies, both of which would break the build for those depending on them. 

Originally my plan was to work on 2.0 as a .NET Core library and leave 1.0 with its legacy" .NET platform targeting, but a 2.0 release (with the new APIs) is too far off.

So, I propose (spurred on by @tpluscode) to do some work now to get the existing DNR APIs built to target .NET Core (probably targeting the netstandard 1.0 and 1.4 platforms). 

So, should this release be numbered 1.1 or 2.0 ? My thought is that 2.0 would be a better choice because we know it is going to break some builds just because of the change in platform targetting. It also gives us the opportunity to look at some of the existing issues such as #79 where there are proposals to change existing APIs. Then that would mean that the full API redesign would be slated for a future 3.0 release instead and so would leave us with:

1.0.x = legacy platform targeting (.NET 3.5, .NET 4.0, PCL, Xamarin) & signed with the old key.
2.x = .NET Core build (netstandard 1.0 and 1.4), some minor API tweaks and breakages but for the most part API-compatible with 1.0.x
3.x = DNR-vNext with completely reworked APIS

How does that sound? @tpluscode, @rvesse any thoughts?

tpluscode:
I think I agree. Cleaning up the build may or may not lead to actual breaking changes although going straight to 2.x would send a clear message that things have changes substantially. The opportunity to do some actually breaking but minor changes could also be good.

The only thing I'm worried about is merging any of those changes (both in build and library) into the future 3.x branch. I suspect that some of that work would have to be done twice.

Finally, do you propose supporting both 1.0.x and 2.x in the long run?

kal:
[...]

Issue #76, opened 2017-02-24
Move to .NET Core/Standard


tpluscode:
@kal, I see that you created the [feature-dotnetcore-build](https://github.com/dotnetrdf/dotnetrdf/tree/feature-dotnetcore-build) branch. The NuGet packages also already come with .NET Standard builds although I don't see any `dotnet`-related stuff in the repo.

How would you feel about attempting to migrate to `project.json`. I know it's being replaced be new `csproj` but personally I've had a hard time getting VS2017 to work. Until it's ready and stable how about we try moving on. It should simplify to repository a great deal where we currently have the project sync and multiple csprojs for each target.

tpluscode:
Oh, my bad. I completely missed the netcore folder. With `xproj`/`project.json` we should be able to have only a single project.

tpluscode:
I've successfully converted a good portion of the libraries and unit tests. A problem I've encountered and don't how to overcome is wrt running tests against portable build. It seems it's not possible with current tooling, because `dotnet` test runner depends on 3rd party packages which don't target PCL.

@kal do you think we have to support PCL or maybe it would be possible to do down to .NETStandard 1.0 or 1.1 instead which cover most of the same platforms?

kal:
IIRC there is a trade-off between functionality and platform support here. 

If possible I would be in favour of just building .NETStandard 1.0 and .NETStandard1.4 targets (with the former being probably as constrained in its features as our current portable library build is), and dropping all of the other variants (including the .NET 3.5 library support). This would possibly break builds for people still trying to support those older .NET platforms - but I'm not sure if there is anyone out there with that sort of hard requirement.

kal:
@tpluscode - Can you commit what you have done so far as a separate branch. Then perhaps we can look at simplifying things further and at issues like PCL support.

tpluscode:
Sure, I'll work some more on this to the point that at least libraries and tests work and I'll push a branch. And net35 isn't a problem unless we'd want to go for 4.5+

kal:
Cool! 🍾 

I would like us to also look at splitting out all the non-core libraries (SPIN, FullText, and the tools) into their own separate GitHub projects and let them version at their own speed and in that way we can further simplify the  core library repository. So if you can just get to the point where the core libraries and their tests are working that would be really useful.

tpluscode:
> their own separate GitHub projects

Agreed! I was thinking the same but about GUI tools.

Issue #75, opened 2017-02-24
Bring back CI


tpluscode:
Back in the day we used my TeamCity too build dnr and run tests. @kal, are you now only running that stuff locally?

I think we should move forward and set up Appveyor. I remember that some tests would fail intermittently (such that relied on dbpedia, etc). Some of them also took quite long. We'll have ignore those tests from running on CI because Appveyor has a limit of 30 minutes for build

kal:
Yep, currently I just run the tests locally. 

A little while ago I got the BrightstarDB build running on AppVeyor, but it is currently broken again because they made some changes to their Xamarin support :-(. However, I like the system and its integration into GitHub and I think it makes sense to do the work necessary to get a build working. I'd be up for doing this, perhaps once we have completed work on #76

tpluscode:
I guess that we could start experimenting on that as soon as I push a half-done branch for #76. Should be later this week.

kal:
Yes definitely!

PR #84, opened 2017-03-04
Building entire solution with dotnet

### LINKED ISSUE #2: Repository restructuring
The main idea with this proposal is to make it easier to work with the source code repo by moving some of the "subprojects" into their own repository. I think it might make sense to do this in stages, start with the really obvious things that can be easily split out from the main repo and then work with that for a bit and see if it makes sense to further split the main repo or the other repos. So I would propose that we end up with the following repos:
- dotnetrdf - the core libraries and all of their related tests (this is Libraries/core and Testing in the original repo)
- samples - sample applications; the contents of the Samples directory in the original repo
- utilities - the rdf tools such as the editor, sparql gui and so on; the contents of the Utilities directory in the original repo plus the editor and editor.wpf projects from the Libraries directory in the original repo.
- virtuoso - the Virtuoso data library
- fulltext - the fulltext indexing library
- spin - the SPIN library

Where references between the different repos is necessary, this can be done using Git submodules.

See also #3

tpluscode:
This PR is to address issues #76, #75 as a side-effect, and partially #82. When complete, it should pretty much resolve #2.

There is buttload of changes, so here's what I've already done and what I have planned for this pull request:

* [x] remove csproj-per target with single, which builds for all platforms
* [x] build system
  * [x] remove remnants of NAnt
  * [x] remove project sync
  * [x] add versioning (I already mentioned GitVersion)
  * [x] make sure signing works
  * [x] replace `.nuspec` with configuration in `*.csproj`
* [ ] running tests
  * [x] convert NUnit to xUnit
  * [x] tests runnable with `dotnet test`
  * [x] tests runnable with VS. don't know why neither ReSharper, nor VS Test Explorer works
  * [ ] running tests for portable build (probably won't be possible unless we move to Standard 1.1 or so)
* [ ] hook up CI
  * [x] create appveyor config
  * [x] running only fast unit tests and such that don't rely on network
  * [x] report unit tests in CI builds
  * [ ] set up NuGet environment


tpluscode:
@kal can you please add an Appveyor web hook to the repository?

kal:
@tpluscode - there is a webhook alreaady there in the settings. I'm guessing AppVeyor added it automatically when you added it as a project (or maybe when I did).

Whats the best way to arrange access to the AppVeyor project? Should I create a dotnetrdf team on my AppVeyor account?

tpluscode:
Please do. I was looking for a way to create a dotNetRDF team but couldn't find it. I use tomasz@t-code.pl on Appveyor

kal:
You should be added as an admin now. I haven't added any projects to the
account for now. I'll let you have that fun with the dotnetrdf project.
I'll look at adding VDS.common later...

tpluscode:
Would you mind me renaming it simply dotNetRDF?

kal:
Do you mean renaming the appveyor account? You can try but I think there is
one that it clashes with...

tpluscode:
How curious. Did @rvesse ever set one up?

tpluscode:
About that webhook, apparently it's not working. Me pushing to the branch didn't trigger a build. I'm guessing because I'm not an admin on GitHub. Do you see a way to add one manually?

tpluscode:
wrt. `nuspec`. Do we want to keep the readme file displayed when installing?

kal:
@tpluscode I've updated the webhook, using the URL provided in the AppVeyor project configuration. Can you try a push and see if it works for you now?

wrt nuspec and readme - I have no strong feeling for keeping the readme displaying when you install the NuGet package. If it is a problem for the project.json, then I say just drop it.

tpluscode:
webhook works :+1: 

The readme is not really a problem. I noticed however that it's probably a little outdated. Just another item to keep track of which personalny is just an annoying tab which opens after installing a package

kal:
I'd be happy to get rid of the readme from the nuget package. As you say, it is one more thing to forget to update...
Lets kill it - we can always bring it back if it turns out that everyone really loved that annoying tab :-)

kal:
Now that the net core tools are updated, I think it would be good to migrate to .csproj as part of this work - what do you think @tpluscode ?

tpluscode:
Will definitely give that a try although on my personal laptop I've been having issues with VS2017 RC so it may be bumpy for me. Unless JetBrains Rider supports it.

tpluscode:
Good grief! They may have made dotnet RTM but 3rd party vendors are struggling. I tried updating to the new tooling and now NUnit isn't supported (see nunit/dotnet-test-nunit#108). On the good side, it seems quite pleasant to work with VS Code but not being able to run tests yet is a PITA.

kal:
@tpluscode hmmm - no unit tests is not good. Did you try the workaround posted on nunit/dotnet-test-nunit#108?
I've tried some porting locally using VS2017 and hit multiple issues so I haven't got as far as getting unit tests working yet, but I don't want to commit that because (a) its a mess! and (b) I don't want to have massive conflicts with the work you have been doing. 

So, it would be good if you could commit what you have done up to now. You could either commit on to this branch which we can just keep working until either NUnit is updated or one of us gets the energy to move the tests to xUnit ;-) or if you prefer, maybe start a separate branch. Given that the goal is really to get all of this going in VS 2017/VS Code I think it would be worth just committing to this branch knowing that for now it will break our unit tests. At least with something committed I might be able to help out with some of the porting work if I get some free time.

tpluscode:
xUnit is a possibility. I did try starting that yesterday but decided it was too much. However I still prefer that (or wait till NUnit works) rather than the wonky workaround. Reminds me of how we'd run tests years back when VS Express couldn't and we hooked up `nunit.console.exe` from a console project. Yuk!

Anyways, I'll push the migrated csprojs. Do you have VS2017 set up already? It installed fine on my work computer so I can continue there.

kal:
Yeah the migration to xUnit would be painful, but I did it a while ago on another project and in the long run I think it would be the way to go. I could take on that pain though :-)

I have VS2017 all set up and ready to go!

Libraries/dotNetRDF/dotNetRDF.csproj
```diff
@@ -64,6 +66,7 @@
   </ItemGroup>
 
   <ItemGroup Condition=" '$(TargetFramework)' == 'net40-client' ">
+    <PackageReference Include="Newtonsoft.Json" Version="8.0.3" />
```
tpluscode: Why different versions of Json.Net?

kal:
9.0.1 doesn't support net40-client. But actually I removed that target
framework so this section can come out.

tpluscode:
You think that we should drop client profile targets too?

tpluscode:
> 9.0.1 doesn't support net40-client

And actually, are you sure? With project.json everything built fine with 9.0.1 in all targets...

kal:
I'll check, but it wasn't listed as a supported framework on nuget...and VS
was refusing to build the net40-client with an error message that indicated
that this was the problem.

I'm ambivalent about the client frameworks. If they build with minimal fuss
then that's all cool, but if they are a PITA to include I would rather drop
them and spend my energy elsewhere!

On 15 Mar 2017 8:57 a.m., "Tomasz Pluskiewicz" <notifications@github.com>
wrote:

9.0.1 doesn't support net40-client

And actually, are you sure? With project.json everything built fine with
9.0.1 in all targets...

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/dotnetrdf/dotnetrdf/pull/84#issuecomment-286679149>,
or mute
the thread

tpluscode:
I don't disagree. I'm just surprised ;). I wonder if people actually used the client profile builds...

kal:
For the sake of sanity and moving things forwards, I've trimmed the target frameworks down to netstandard1.4, net40 and net40-client. So I've ditched net35, net35-client and the PCL build.

kal:
One other thing I've noted is that the netstandard1.4 has compiler flags that exclude quite a lot of functionality. I originally put these in as a quick and dirty fix for the lack of compatible APIs in the target framework. It would be good to see if we can find a way to add back the support. The flags currently defined are:

- [ ] NO_SYSTEMCONFIGURATION - this indicates that the System.Configuration.ConfigurationManager API is not available and disables support for the ConfigurationLoader. We could maybe add a NetcoreConfigurationLoader to replace this functionality using the .NET Core configuration APIs
- [ ] NO_PROXY - this indicates that the System.Net.WebProxy API is not available and disables proxy support in a number of HTTP endpoint implementations.
- [ ] NO_XSL - this indicates that the System.Xml.Xsl namespace is not supported. It disables support for XSL stylesheets in the TriX parser and disables the SparqlXsltWriter. I'm not sure there is anything we can do about this one.
- [ ] NO_HTMLAGILITYPACK - indicates that the HtmlAgilityPack library is not available. This means no RDFa parser amongst other things :-(
- [ ] NO_DATA - this indicates that the System.Data.DataTable API is not available. It just disables the APIs that allow graphs and SPARQL results to be acccesed as a DataTable, so not really a problem I think.
- [ ] NO_WEB - this has a huge impact in terms of supported features and needs a bit of careful review because I think that in many places we do have access to the required APIs. Needs further investigation.
- [ ] NO_PROCESS - indicates no support for starting external processes (ProcessStartInfo API?). Only used by the GraphVizGenerator class.
- [ ] NO_URICACHE - indicates that the VDS.RDF.Parsing.UriLoaderCache class is not included and disables the use of that cache in the UriLoader class.


I think it would be worth going through each one and making a call on whether it stays in as a compiler flag or we do something to work around it for this target profile.

tpluscode:
wrt to HtmlAgilityPack, we could replace it with [AngleSharp](https://www.nuget.org/packages/AngleSharp/). It's new, shiny and supports it all down to PCL and Standard 1.0

tpluscode:
> we could replace it with AngleSharp
that said I'd create separate tasks for this and cleaning up the compiler symbols

kal:
I agree. I'm in the middle of porting the tests to xunit so we can
reinstate that bit if the build.

tpluscode:
Not sure why, but the unit test project fails to restore via the `sln` file:

> warning MSB4121: The project configuration for project "unittest" was not specified in the solution file for the solution configuration "Release|Mixed Platforms"

Not sure where it's coming from. I'm using VS Code and it doesn't work with solutions I think. While you work on xUnit, can you please look into that?

kal:
I'll take a look - which sln file are you trying to build ? I'm currently using dotNetRDF.NetCore.sln as it contains the minimal set of projects.

tpluscode:
I was using the "full" one. So maybe if you change the `dotnet restore` line in `appveyor.yml` it will do the trick?

kal:
The restore is working now and it looks like stuff is compiling. I think the appveyor build now needs some tweaking.

tpluscode:
Build now works and creates packages. Did you manage to finish migrating to xUnit? I tried running locally and got

> Starting test execution, please wait...
[xUnit.net 00:00:01.1177369] Skipping: dotNetRDF.Test (Could not find any of the following assemblies: xunit.execution.desktop.dll)
No test is available in E:\github\dotnetrdf\Testing\unittest\bin\Debug\net451\dotNetRDF.Test.dll. Make sure that installed test discoverers & executors, platform & framework version settings are appropriate and try again.

kal:
Just working on that now - its another target framework issue!

kal:
There are still a few test failures.

A bunch of tests are failing because I messed up one of the global search/replace regexes :-( - I'll just go through and manually fix them.

Another group of tests are failing because of a missing configuration file on the dotnetrdf.org server. A quick workaround for now will be to skip those files while I try and figure out what that file was! @rvesse if you have any pointers to the config file in the repository that would be great.

It looks like there may also be a few failures due to not finding an embedded resource or a content file, but I'll have to dig a bit deeper into that.

tpluscode:
I'm seeing a lot of

> VDS.RDF.XunitExtensions.SkipTestException : UnitTestConfig.properties cannot be found, to configure your test environment please make a copy of UnitTestConfig.template under the resources directory, add it to this project as a Content item and then edit it to match your test environment

and apparently the build hangs. The same happens to me locally. At some point I get `SkipTestException` and the is no more console output. Are you running the tests fine with VS?

kal:
Tests run fine in VS. Some tests run for a long time though, but they do
complete.

SkipTestException is the replacement for Asset.Inconclusive, and the test
that throws it should have a SkippableFact attribute on it (I may have
missed it in some places because its not always obvious).

tpluscode:
Funny how the test report on AppVeyor is inconsistent. Or is it that the test runs produce different results. I think I've seen both cases reported on AppVeyor's support page.

In any case, I see that most failed tests come from `VDS.RDF.Query`. Do you have an idea if there's a common cause? 

Other than that there's `ConfigurationImports4`, which I think passed in the previous build? 

Finally, `ParsingRdfXmlWithUrlEscapedNodes2` stands out, which is kind of weird. Without going into why the comparison fails, I'd say that assertion doesn't belong to that test case anyway. I would remove it and investigate separately if reported.

kal:
Smells like a parallelism issue. I spotted that several test classes are using the path "temp.ttl" or "resources\\temp.ttl" for their test data and as XUnit runs different test classes in parallel there is a chance that these classes will delete/overwrite each other's test data. I think the solution here is to use different temp file names in each class, so we can keep the parallelism.

A similar problem afflicts the Lucene tests which all share the same static index instance across several test classes. Here I think the best solution is to put all of the test classes into a single collection to force xUnit to run them serially with respect to each other.

tpluscode:
`dotnet test` still hangs at some point when running `ExplainProcessorTests` :(

kal:
I'm close to giving up on those full text tests. They shouldn't be running in parallel now, they work just fine on my local machine and I have no idea why they sometimes all pass on the build server and sometimes fail.

My only remaining guess is that there is some persistent state (again possibly something in the Options class) that another test is setting that causes these tests to fail. I noticed that if Options.AlgebraOptimisation is disabled, then the full text query processing doesn't happen (because its implemented as an optimisation), but I've modified the tests to explicitly set this option right before executing the query and still no joy...)

I'm going to try running the full text query tests in a separate run in the appveyor configuration and see if that helps, but if it doesn't I'm out of ideas :-(

kal:
Well, that didn't work either because it looks like the category filtering doesn't work with to != So I've reverted that change. I am now totally out of ideas. One possible (desperate) solution would be to move all the full text unit tests into a separate assembly (which would sort of make sense as the full text assembly is itself separate from the core assembly) - but if the issue is to do with the platform the unit tests are running on, then that won't help.

@tpluscode - any thoughts on how we should proceed from here?

tpluscode:
Whenever I run all tests (ReSharper) I consistently get 15-16 failures. `FullTextGraphScopingTests`, `FullTextOptimiserTests` and one or two tests from `QueryTimeouts`. 

Weird with these filters. I have another idea though. See latest commit

kal:
Nice idea! Let's see what happens...

tpluscode:
Bugger, builds fine locally. Let me fix that

tpluscode:
uh, oh. build successful but one failed test

kal:
Awesome work, thanks Tomasz!

I wonder if we should roll the tools into the build now or not. I would prefer to split them out into a separate project that consumes the core libraries via NuGet. If we agree on doing that then perhaps we should move on to creating a separate 'tools' repo (as a clone of this one?) and then we can continue tidying this repo up to remove the old NAnt build etc.

What do you think?

tpluscode:
Yes, I'd make the toolkit into a separate repo too. In fact I'll do it first thing. Not sure about samples. We could keep the versioning in sync on the major.minor level and let the lower number diverge. Shouldn't be a problem.

kal:
Currently the samples are not in any distribution (as far as I know), so letting them just exist as a separate repository seems to me like a good plan. We can tag with major.minor version and with each minor/major release we just update the NuGet references and make any changes required by breaking API changes. Other than maybe having an appveyor build to validate that everything compiles correctly against the referenced DNR version I think we don't need anything else for the samples repository.

tpluscode:
Kal, are you extracting the samples folder? If not I'll do it and preserve history.

Also, please create an environment on AppVeyor for NuGet and GitHub release. I think only you can get the API key for nuget and access token for GitHub?

kal:
Hi Tomasz, I'm traveling today so I won't get chance to extract the samples
today. If you could do it that would be great. I think the benchmarking
stuff could also go into the samples repo.

tpluscode:
In fact I did, shortly before you wrote. By benchmarking stuff you mean projects under `Testing` other than the uni tests project?

By the way, the timeout tests seem flaky. You have an idea how to fix them?

kal:
I was meaning just Testing/GraphBenchmarker. Is it possible to do Git magic to copy those files over with history? If not, I'll do it the cut-n-paste non-magic way :-)

Some of the query timeout test cases make assumptions about how long operations are taking, I think that occassionally we may just hit a case where a query gets executed faster than the timeout. For the most recent failure I've just dropped the query timeout threshold, but I think we need to look again at exactly what those tests are testing and figure out if there is a way to mock the rest of the execution so that it will always take longer than the timeout threshold.

tpluscode:
> Is it possible to do Git magic to copy those files over with history?

It should be possible. I'll try

kal:
I've added skeleton deployment sections for NuGet and GitHub with the API keys required for those method. I've set them both to only deploy from master. If we go for the "PR into master" approach for the repository, we may want to also set some additional rules for publishing releases...although I would be happy to experiment with simply publishing a new release each time we update master.

tpluscode:
Pushing automatically on master would mean that we keep develop, right? 

Also, let's discuss versions. Currently every build on master would get the same "stable" version, which is calculated as patch-increment of last tag stable tag. Current tag is `1.0.13-pre7` and so master gets version `1.0.13` which is the next. The thing is that each subsequent build will get the same number, which means you won't be able to update nuget.org. The only way is to tag master. This is called `ContinuousDelivery` in GitVersion and is the default behavior

The other way, called `ContinuousDeployment`, adds a prerelease tag to every build after tag appended with number of commits since that tag. Default is `-ci`. Currently the version would be `1.0.13-alpha72`. This is somewhat suitable provided we don't spam master with too many commits as I don't think we want a whole lot of prerelease packages on NuGet.

To sum up, I think we have three possible workflows:

1. No develop - `ContinuousDelivery`
    * Push everything to master
    * Tag next version
    * Wait for tagged commit to build
    * Push to nuget manually on AppVeyor
       * should be possible to push tags automatically, but [AppVeyor is weird in that regard](http://help.appveyor.com/discussions/questions/1810-deployment-to-github)
1. Develop & master - `ContinuousDelivery`
    * Accept PRs only to develop
    * When version is ready, merge to master 
        * possible release branch
    * Have AppVeyor build and push
    * Tag master whenever, because the version is the same for all commits between tags
1. Develop & master - `ContinuousDeployment`
    * Accept PRs to develop or master
    * When ready, merge to master 
    * Tag to have next stable release pushed

What do you think?

kal:
I'm willing to experiment with (1) because it feels like it will be a simpler and more streamlined process. Although I think I would prefer to still use PRs to update master rather than just push updates directly (especially with having Appveyor run the unit tests on the pull request as a useful smoke test for the update). The "pain" of having to manually force a nuget deployment from Appveyor isn't really an issue for me...especially when compared to our current deployment process!

tpluscode:
Definitely. I too send PRs to my own repositories to only merge when tests pass and code coverage is satisfactory.

kal:
Just to be sure I understand the ContinuousDelivery approach. Right now the last tag on master is `1.0.13-pre7`, so a merge of this branch into master would trigger a `1.0.13-pre7+1`. If we were then happy to release this as our `2.0.0-pre1` release we could tag master as `2.0.0-pre1` and when that tag gets pushed, Appveyor will do the build and (assuming the build succeeds) deploy a `2.0.0-pre1+0` release on GitHub (and if we want to we can push a `2.0.0-pre1` deployment to NuGet). And we just continue like that, tagging master with a SemVer for each release we want to deploy.

Is that correct? I hope so, because it sounds all good to me!

I've cleaned out the old NAnt and mercurial scripts from Build, so you can cross those items off the checklist. It might be good to get the Sandcastle documentation built as another artifact that we can include in the GitHub release, so maybe I'll take a quick look at adding that into the Appveyor script.

tpluscode:
Almost correct. Merging to master a well as any other commit would trigger `1.0.13` and it will stay until you tag again (or bump the version with commit message). The [documentation](https://gitversion.readthedocs.io/en/latest/more-info/version-increments/) explains why you don't get a new version with every commit by default.

> Semantic Versioning is all about releases, not builds. [...] All this effectively means that GitVersion will produce the same version NuGet package each commit until you tag a release.

You may want to read the page linked above to learn about additional options. But in general, yes, we would tag master with every release and publish those builds.

kal:
Yeah, I read that documentation and I'm still a bit confused about it, maybe its just really simple and I'm over-thinking it!

I think we are ready to merge this PR. I've added a short entry to the change log and bumped the major version in the commit message, so a merge into master will result in a 2.0.0 build - hope that's right! :-)

tpluscode:
GitVersion takes some getting used to and it is quite intimidating indeed ;)

Shall I change this PR to merge to master instead of develop then?

Also, nice work getting Sandcastle to build too. How are we publishing the docs? I'd probably add the chm to GitHub release and what about the zipped HTML version?

kal:
I think it would be worth adding both the zip and the chm to the GH
release. If you could merge this PR into master that would be great... I'm
going to be AFK for a while...

kal:
PR merged. I've also updated the repository settings so that the default branch is now master.

tpluscode:
I think we forgot last thing about the benchmark projects. I won't have the time today and likely hangover tomorrow. Please go ahead with just copying the non-magical way if you have a second.

PR #89, opened 2017-04-01
Removal of NO_WEB #def from netstandard target

kal:
This is a PR to add the web functionality back in to the netstandard1.4 build of the core library.

The main pieces of this are:

- [x] Remove the NO_WEB #def and fix any resulting compilation issues
- [ ] Reimplement the custom ASP.NET HTTP Handlers in VDS.RDF.Web as ASP.NET 5 Middleware

PR #92, opened 2017-04-08
Add RDFa parser to netstandard target

kal:
This PR replaces #89.
The former PR meant that net40-client target framework had to be removed (not supported by AngleSharp) and there were some nasty app.config hacks needed to compile the net40 build (due to an upstream dependency of AngleSharp).

This PR takes an alternative approach so that instead of replacing HtmlAgilityPack for the net40 and net40-client frameworks, we only use it for implementing the RDFa parser for the netstandard1.4 framework.

This has the upside of allowing us to continue supporting net40-client and the small downside of requiring some separate unit testing for the netstandard framework.

This PR does not address the implementation of ASP.NET middleware for the various web protocols.


kal:
@tpluscode - what do you think. I'm in two minds about this approach. If we go with just AngleSharp, then net40-client has to be dropped and net40 requires app.config changes to fix a borked Microsoft.Bcl nuget package dependency. On the other hand this PR means we have a parallel code base for the RDFa parser. I suppose a better approach would be to refactor the RDFa parser so that there is a common HTML parser interface, but that just feels like more work :-(

tpluscode:
Given what you describe about net40(-client) support, I'd stay with HtmlAgilityPack for that target. If there is possibility to share some base code and differentiate only between the implementation details, I think it's worth a shot. I'll have a look that.

Additionally I have two other comments:
*  I would not create a separate test project because dependencies. `dotnet test` can handle multiple targets and the inclusion of source files can be done in `csproj`. It's ironic though that I quickly discovered that with multi-target test project we'll need a separate project for Query.FullText and Data.Virtuoso. What do you think?
* And so I would try to limit the use of precompiler directives in favor of said inclusion/exclusion of whole `cs` files wherever possible. Would you accept partial source files, separate for targets for example? Say, something like `*.Netstandard.cs` and `*.Net40.cs` wildcards.

tpluscode:
I have some changes under way. Let's see if you'll like them :)

kal:
@tpluscode - I agree with both points above. Because DNR predates the partial class construct there is a lot of code using #defines. Replacing them all will be quite a bit of work, but it would also be an opportunity for review and removal of any that are now obsolete (for example sections in #if SILVERLIGHT are probably unused now) and just some general tidying up. However I would say for right now lets just focus on the RDFa parser - getting it back into the .NET Core build would be a nice step forward.

tpluscode:
I know you're right. However I'm pretty much finished with some changes which address what I summarized yesterday. It turns out that in many cases the changes required are all or nothing and so it may seem substantial. Not complete though, just enough to get a multi-target unit tests project running. I haven't touched any of your code which actually addresses the issue. Just spitting some stuff which didn't build on both platforms... A good starting point for getting all that stuff under control I think :)

I'll add a separate ticket for a thorough cleanup.

tpluscode:
> I'll add a separate ticket for a thorough cleanup.

Oh, wait, what was the plan regarding the old codebase vs the 1.9 branch?

tpluscode:
So, here are my changes. Significant and exceeds the scope of the PR but now we run all tests possible against .NET Core. 

More interestingly though I probably payed off after. After eliminating all tests which could not run on .NET Core framework there are two tests which fail on parsing RDFa only in .NET Standard build. I'll still take some time adjusting the Appveyor script but do check them locally if you have the time

tpluscode:
Hm, I tried upgrading the xunit runner (now comes with `dotnet test` command and better support in general). Unfortunately it caused the optimizer and some SPARQL tests to fail on both platform. I think you were fixing them before. Would you know what's wrong?

The failing RDFa tests are `GraphHandlerTests#ParsingGraphHandlerExplicitRdfA` and `WriterTests#WritingHtmlWriter`. Both seem to get the graph's `BaseUri` wrong

Libraries/dotNetRDF/Parsing/RdfAParser.NetCore.cs
```diff
@@ -68,7 +68,7 @@ protected override IHtmlDocument LoadAndParse(TextReader input)
 
         protected override bool HasAttribute(IElement element, string attributeName)
         {
-            return !element.HasAttribute(attributeName);
+            return element.HasAttribute(attributeName);
```
tpluscode: Good catch. I was wondering why the GoodRelations test started failing all of the sudden

kal:
Looks like those couple of small fixes have it sorted out now. The separation of the RDFa parser code is a lot cleaner too - thanks @tpluscode !

tpluscode:
Cool, I didn't think that the last fix would be that simple :)

There is jus one gotcha with the current change to the `*.csproj` files which target multiple platforms. See how I added `<EnableDefaultCompileItems>false</EnableDefaultCompileItems>` and replaced that with lines similar to

```
<Compile Include="**\*.cs" Exclude="obj\**\*.cs;**\*.NetFull.cs" />
```
and 

```
<Compile Include="**\*.cs" Exclude="obj\**\*.cs;**\*.NetCore.cs" />
```

It's the best way I found to have a separate set of files for Core and Framework. I tried with `<Compile Remove="" />` but adding that to any conditional section removes those files from Solution Explorer completely which is kind of pointless.

Anyway, this works but there is a weird side effect where sometimes when you add or rename a file it tends to be added explicitly with `<Compile Include=""/>` and so MSBuild wants to build it for all targets. We'll have to remember nad maybe ask around for a better solution.

kal:
OK, that will be a good thing to remember as we tidy up the compile flags and so on, because I imagine there will be quite a lot of file adding/renaming to do!

PR #97, opened 2017-04-23
CORE-442 Compressing turtle fix

### LINKED ISSUE #10: Compressing Turtle Writer fails to properly compress RDF Collections and Base Uris
*Imported from http://www.dotnetrdf.org/tracker/Issues/IssueDetail.aspx?id=442*

The Compressing Turtle Writer does not correctly compress RDF Collections or Graph base URIs. An RDF Collection with one item in the list does compress, however, if there are more than 1 items in the RDF Collection then bnodes are rendered in the compressed turtle. When a base URI is set on the graph, the @base directive is written. However, the turtle is not compressed.

Example:

    <urn:test> a <urn:Test> ;
                <urn:hasTest> ( <urn:test/id/1>
                                <urn:test/id/2> ) .

when ran through the compressing turtle writer at high compression

    @prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>.
    @prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>.
    @prefix xsd: <http://www.w3.org/2001/XMLSchema#>.
    
    _:autos11 rdf:first <urn:test/id/1>;
              rdf:rest _:autos12.
    _:autos12 rdf:first <urn:test/id/2>;
              rdf:rest rdf:nil.
    <urn:test> a <urn:Test>;
               <urn:hasTest> _:autos11.

When there is only 1 item in the rdf collection, the turtle generates correctly. However, it fails if there are multiple.
For the @base issue...

    @base <http://somebaseuri/> .
    
    <test/id/1234> a <urn:Test> ;
                   <urn:hasTest> ( <test/id/321> ) .

When ran through the writer at high compression, the @base is generated. however, the uri nodes are not compressed.

    @base <http://somebaseuri/> .

    <test/id/1234> a <urn:Test> ;
                   <urn:hasTest> ( <test/id/321> ) .

result

    @base <http://somebaseuri/>.

    @prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>.
    @prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>.
    @prefix xsd: <http://www.w3.org/2001/XMLSchema#>.

    <http://somebaseuri/test/id/1234> a <urn:Test>;
                                      <urn:hasTest> (<http://somebaseuri/test/id/321>).

---

Rob Vesse (Tuesday, April 07, 2015 4:50 PM) 

We never actually use the Base URI for compression since that would generate relative URIs which while more compact are considerably more fragile.

dotNetRDF always favours correct output over compressed output so in the collection case it is perhaps being overly cautious.  Though this may also be an interaction with another feature of the writer where by it falls back to more verbose output if it thinks the graph won't compress well.

---

Zandt Tittle (Wednesday, April 15, 2015 1:25 AM)

I can understand the complexities faced when trying to compact the graph base uri. What I don't understand is if its not supported, then why is the @base directive being generated? It's a little misleading and adds superfulous information if its not really being used in compaction.

As far as the list handling, I fail to see how this should be treated as 'overly cautious'. Either list compression is supported or it is not. It's not clear to me how this would effect round tripping of the data unless special care is being given to treatment of bnodes. Perhaps an experimental compression option would be useful if there is a slight chance of data loss? 

---

Rob Vesse (Monday, April 20, 2015 5:31 AM)

The Base URI is output because it is information associated with the graph, in some systems that information is used as an identifer for the graph and it is a requirement that serialization and deserialization be lossless so even if we don't otherwise use it for compression we will always output it if the serialization format supports it.

"Overly cautious" was a generalisation about serialization as a whole, we have always preferred correctness over compactness and will continue to do so.  Clearly your test case demonstrates a case where this caution goes too far.  There are a lot of nasty things that people can do with RDF collections and blank nodes that need to be accounted for e.g. non-references, multiple references, concrete list nodes, circular references, interdepent collections etc

From debugging I can see that the rountines for collection compression correctly identify the collection but then incorrectly mark it as unsafe for compression which as you point out is clearly a bug which we will look to fix in the next release.

tpluscode:
Work started on old issue CORE-442

Fixes #10

### DIFF
--- a/Libraries/core/net40/Writing/WriterUtilities.cs
+++ b/Libraries/core/net40/Writing/WriterUtilities.cs
@@ -566,12 +566,14 @@ namespace VDS.RDF.Writing
                 }
                 else
                 {
+                    //CORE-442 This calculation may be incorrect
+
                     //For implicit collections if the number of Triples in the Collection is exactly ((t*3) - 1) those in the Graph then
                     //can't compress i.e. the collection is not linked to anything else
                     //Or if the number of mentions compared to the expected mentions differs by more than 1 then
                     //can't compress i.e. the collection is linked to more than one thing
                     int mentions = context.Graph.GetTriples(kvp.Key).Count();
-                    int expectedMentions = ((c.Triples.Count * 3) - 1);
+                    int expectedMentions = 2;
                     if (expectedMentions == mentions || mentions-expectedMentions != 1)
                     {
                         context.Collections.Remove(kvp.Key);
--- a/Testing/unittest/Writing/CollectionCompressionTests.cs
+++ b/Testing/unittest/Writing/CollectionCompressionTests.cs
@@ -497,7 +497,7 @@ namespace VDS.RDF.Writing
         }
 
         [Test]
-        public void WritingCollectionCompressionCore442_03()
+        public void WritingCollectionCompressionCore442_02()
         {
             Graph g = new Graph();
             g.LoadFromFile(@"resources\core-442\three-items.ttl");
--- a/Testing/unittest/Writing/CollectionCompressionTests.cs
+++ b/Testing/unittest/Writing/CollectionCompressionTests.cs
@@ -467,5 +467,48 @@ namespace VDS.RDF.Writing
 
             this.CheckCompressionRoundTrip(g);
         }
+
+#if !NO_FILE
+
+        [Test]
+        public void WritingCollectionCompressionCore442_01()
+        {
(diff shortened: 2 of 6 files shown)


tpluscode:
@kal I was looking over some stuff on my GH account and found this. I'm pretty sure it's outdated and we can close it...

kal:
Yeah it is a hold-over from the migration to GitHub. The branch is still there but I think we can close the PR.

PR #98, opened 2017-04-23
Old Fluent query branch

tpluscode:
Cleaning up branches. Let's see if there are any outstanding changes and close this branch


kal:
It looks like the useful bits of this branch are already merged. @tpluscode : can we just close this PR?

tpluscode:
I think so.

PR #99, opened 2017-04-23
An OWL API for .Net based on Clark & Parsia's Ortiz API

tpluscode:
Unfinished work on OWL2 API.

- [ ] Adapt to .NET Core project structure


rvesse:
This appears to be simply a recovery of abandonware stub from an old abandoned branch. It provides no value to users and requires substantial work to reach a stage where it would be usable. There are already more features than the community is able to test and maintain so embarking on creating additional ones seems low priority to me.

tpluscode:
Yes, I wanted to clean up old branches and wasn't sure what was the state of this. No harm in keeping it here anyways

kal:
I think it is unlikely to get any further work on it. I'm happy to leave the branch, but I suggest that we close the PR - what do you think @tpluscode, @rvesse ?

tpluscode:
I won't object. Please go ahead.

PR #109, opened 2017-06-11
[StardogConnector] Query with reasoning parameter

charbull:
Updated the query function to add the reasoning parameter activation in the query.

In the [Java API](https://www.stardog.com/docs/java/snarl/com/complexible/stardog/api/query) it is possible to activate the reasoning by query:
```java
Query<T>	reasoning(boolean theReasoningFlag)
Specify whether you would like the query to be executed with reasoning.
```

This pull request adds two functions
```C#
Query( String sparqlQuery, bool reasoning).
```
and 
```C#
public virtual void Query(IRdfHandler rdfHandler, ISparqlResultsHandler resultsHandler, String sparqlQuery, bool reasoning)
```
It is preferable to have one `Query` function with the reasoning parameter instead of two but this will have impact on `IQueryableStorage`

### DIFF
--- a/Libraries/dotNetRDF/Storage/IStorageProvider.cs
+++ b/Libraries/dotNetRDF/Storage/IStorageProvider.cs
@@ -294,6 +294,40 @@ namespace VDS.RDF.Storage
     }
 
     /// <summary>
+    /// Interface for storage providers which allow SPARQL Queries to be made against them with reasoning set by query
+    /// </summary>
+    public interface IReasoningQueryableStorage
+        : IStorageProvider
+    {
+        /// <summary>
+        /// Makes a SPARQL Query against the underlying store
+        /// </summary>
+        /// <param name="sparqlQuery">SPARQL Query</param>
+        /// <param name="reasoning">rReasoning On demand by query</param>
+        /// <returns><see cref="SparqlResultSet">SparqlResultSet</see> or a <see cref="Graph">Graph</see> depending on the Sparql Query</returns>
+        /// <exception cref="RdfQueryException">Thrown if an error occurs performing the query</exception>
+        /// <exception cref="RdfStorageException">Thrown if an error occurs performing the query</exception>
+        /// <exception cref="RdfParseException">Thrown if the query is invalid when validated by dotNetRDF prior to passing the query request to the store or if the request succeeds but the store returns malformed results</exception>
+        /// <exception cref="RdfParserSelectionException">Thrown if the store returns results in a format dotNetRDF does not understand</exception>
+        Object Query(String sparqlQuery, bool reasoning);
+
+        /// <summary>
+        /// Makes a SPARQL Query against the underlying store processing the resulting Graph/Result Set with a handler of your choice
+        /// </summary>
+        /// <param name="rdfHandler">RDF Handler</param>
+        /// <param name="resultsHandler">SPARQL Results Handler</param>
+        /// <param name="sparqlQuery">SPARQL Query</param>
+        /// <param name="reasoning">rReasoning On demand by query</param>
+        /// <exception cref="RdfQueryException">Thrown if an error occurs performing the query</exception>
+        /// <exception cref="RdfStorageException">Thrown if an error occurs performing the query</exception>
+        /// <exception cref="RdfParseException">Thrown if the query is invalid when validated by dotNetRDF prior to passing the query request to the store or if the request succeeds but the store returns malformed results</exception>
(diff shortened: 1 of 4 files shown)


kal:
Firstly, thanks for the PR!

I wonder if it might be better to change the implementation though to use a boolean flag on the connector instead of extending the API. That way you can continue to use the standard IQueryableStorage APIs.

I'm thinking that the code would be something like:

```
var stardog = new StardogConnector(....);
stardog.ReasoningEnabled=true;
stardog.Query(...);
```

The advantage of doing it this way is that any function receiving the connector as a parameter can just treat it as an IQueryableStorage instance without having to know about special Stardog-specific query methods.

The other question I have is if this is something that has changed in either the V4 or V5 Stardog API. I think it might be because the existing code is definitely set up to send the reasoning mode through on queries - perhaps that was supported in older versions of the Stardog API? Anyway because of this I think it would be better to move the code to a new connector class rather than change the base class which would risk causing problems for folks connecting to an older version of the Stardog server.

Final thing to ask - can you add a unit test for your PR to `Testing\unittest\Storage\StardogTests.cs`

Cheers

Kal

tpluscode:
> The advantage of doing it this way is that any function receiving the connector as a parameter can just treat it as an IQueryableStorage instance

Unless you really need this setting on per-query basis. Is that the intended usage @charbull?

charbull:
@kal I see your point, that's why I didn't want to touch `IQueryableStorage`

@tpluscode as you mentioned, the setting by query would be more convenient. As not all the queries have the inference activated. This means the user will need to set the reasoning to be enabled before the queries.

I see the three following solutions:
1. Almost no modifications for `IQueryableStorage` but the user will need to set the reasoning before each query when it needs to be activated/disactivated.
```C#
var stardog = new StardogConnector(....);
stardog.ReasoningEnabled=true;
stardog.Query(...);

stardog.ReasoningEnabled=false;
stardog.Query(...);

stardog.ReasoningEnabled=true;
stardog.Query(...);
```

2. Modify `IQueryableStorage` by adding a reasoning parameter to the Query method.
```C#
Query( String sparqlQuery, bool reasoningEnabled)
```
or just a dedicated function to query with reasoning, since in this case the boolean will always be true, if not the user will just use the old one Query(String Sparql) and will maintain a previous compatibility.

```C#
QueryWithReasoning( String sparqlQuery)
```

cons: modifying `IQueryableStorage` will impact other connectors.
pros: it is a good solution for the long term since this behavior is not just specific to StardogConnector, AllegroGraph has a reasoning by query as well. So it can be used for the allegroGraph connector as well.  I am not sure about the reasoning situation in Virtuoso?

3. No Modifications to `IQueryableStorage` just add a new function to the StardogConnector. 
```C#
QueryWithReasoning( String sparqlQuery)
or 
Query( String sparqlQuery, bool reasoningEnabled)
```

For the long term I would go with Solution 2, as AllegroGraph will have a similar modification. 
For the short term I have a perference with solution 3 since it requires less code for the user?

What do you think? Other approach?

Cheers,
Charbel

kal:
How about a 4th possible solution :-)

Create another interface IReasoningQueryableStorage and then add the query methods to that. This way we can then add that interface to just those storage connectors that support query time enabling/disabling of reasoning. I think keeping the boolean flag would be more useful for developers because it would allow them to use any boolean expression to disable reasoning instead of having to wrap an if statement around calls to two different query functions. So something like:

```
public interface IReasoningQueryableStorage {
  object Query (string sparqlQuery, bool reasoning);
  void Query(IRdfHandler rdfHandler, ISparqlResultsHandler resultsHandler, String sparqlQuery, bool reasoning)
}
```

This just feels a bit cleaner, keeps the flexibility of having the reasoning flag as a method argument and would enable us to add the functionality to the other connectors in stages.

charbull:
I like it!! It is clean and flexible :)

I am on a trip now. Will update it on Monday if that's ok with you :)

tpluscode:
Definitely options 3 and 4 are best. In fact, I think that adding an overload of `Query` to StardogConnector is pretty much the equivalent of introducing a new interface. Both avoid modifying existing contracts. 

If other connectors really support this, I'd add the interface. Otherwise it could be an overkill but at the same there's no harm in doing so.

charbull:
So we go for option 4?

kal:
Option 4 seems like the best to me.

charbull:
@tpluscode  and @kal pull request updated with option 4.

tpluscode:
LGTM

kal:
Merged. Thanks @charbull !

charbull:
Great thank you !

Issue #111, opened 2017-06-15
SparqlParameterizedString does not recognize default namespace unless it is listed last


christopheralcock:
Strange error I've noticed today is that if you include more than one PREFIX statement in a sparql query string, dotnetrdf denies that you've stated the default (`PREFIX :`) namespace, unless you state it after all other prefixes.
Work-around is easy, but this seems like a very strange bug to me.

kal:
Fixed in PRs #113 and #115

PR #113, opened 2017-06-16
Updates regex to recognise default prefixes

christopheralcock:
solves https://github.com/dotnetrdf/dotnetrdf/issues/111
h/t @langsamu
Current behaviour is that if you only say `PREFIX : <http://example.com#>`
you get the desired behaviour that `:Thing` resolves to <http://example.com#Thing>
However if you say the following:
`PREFIX : <http://example.com#>`
`PREFIX another: <http://anotherexample.com#>`
the first prefix is stripped and `:Thing` will not resolve and the validator says you've not declared the default namespace

Regex fix makes dotnetrdf treat `PREFIX : <http://example.com>` in the same way as `PREFIX example: <http://example.com>`, and lets you state your prefixes in whatever order you like, as per SPARQL's grammar rules: https://www.w3.org/TR/2013/REC-sparql11-query-20130321/#grammar

### DIFF
--- a/Libraries/dotNetRDF/Query/SparqlParameterizedString.cs
+++ b/Libraries/dotNetRDF/Query/SparqlParameterizedString.cs
@@ -94,7 +94,7 @@ namespace VDS.RDF.Query
         private static Regex ValidParameterNamePattern = new Regex("^@?[\\w\\-_]+$", RegexOptions.IgnoreCase);
         private static Regex ValidVariableNamePattern = new Regex("^[?$]?[\\w\\-_]+$", RegexOptions.IgnoreCase);
 
-        private static Regex PreambleCapturePattern = new Regex("(BASE|PREFIX\\s+([\\w\\-_]+):)\\s*<([^>]+)>", RegexOptions.IgnoreCase);
+        private static Regex PreambleCapturePattern = new Regex("(BASE|PREFIX\\s+([\\w\\-_]*):)\\s*<([^>]+)>", RegexOptions.IgnoreCase);
 
         /// <summary>
         /// Creates a new empty parameterized String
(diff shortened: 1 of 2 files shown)


tpluscode:
Hi Christopher. Thank you so much for the fix. Could you also provide a failing test case. A snippet at least.

@kal hm, where would we put that test?

christopheralcock:
Hi @tpluscode 
https://github.com/ukparliament/Data.Api.FixedQuery/blob/master/Parliament.Data.Api.FixedQuery/Controllers/ParliamentController.cs#L46
If lines 46 and 47 were swapped with eachother that would fail, and this fix should make it work in either order

kal:
@tpluscode @christopheralcock - a test added to `Testing\unittest\Query\ParsingTests.cs` would be great.
Thanks for the PR!

tpluscode:
@christopheralcock actually, I see that tests of `SparqlParameterizedString` are mostly in `Testing\unittest\Query\SparqlTests.cs`. Please add the following there:

``` c#

[Fact]
public void SparqlParameterizedStringDefaultPrefixOrder()
{
    // given
    const string queryString = @"
PREFIX : <http://id.ukpds.org/schema/>
PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
CONSTRUCT *
WHERE {
?s ?p ?o
}";

    // when
    var query = new SparqlParameterizedString(queryString);

    // then
    query.Namespaces.Prefixes.Should().HaveCount(2);
    query.Namespaces.GetNamespaceUri(string.Empty).Should().Be(new Uri("http://id.ukpds.org/schema/"));
}
```

kal:
Looks like that broke the build on Appveyor - maybe just a missing import of the FluentAssertions namespace?

```
Query\SparqlTests.cs(712,60): error CS1061: 'Uri' does not contain a definition for 'Should' and no extension method 'Should' accepting a first argument of type 'Uri' could be found (are you missing a using directive or an assembly reference?) [C:\projects\dotnetrdf\Testing\Unittest\unittest.csproj]
```

tpluscode:
Yes, my bad. I reverted that line thinking it was in the other file.

PR #148, opened 2017-12-18
Move HTTP authentication code to base class

tpluscode:
See #149


kal:
Thanks for the PR @tpluscode. As with the other branch I also moved the configuration serialization code to the base class.

markdwood:
BTW, what are the plans around creating a new “stable” NuGet package?  The last one marked stable is 1.0.12.

Also, what is the philosophy around when to release new 2.0 prereleases?

Thanks,
-Mark

PR #149, opened 2017-12-18
Move HTTP authentication code to base class

tpluscode:
This way any connector which inherits from `BaseHttpConnector` will be able to use basic authentication.

I needed this for Fuseki but simply moved up the inheritance tree


kal:
Thanks @tpluscode - I merged with one change to also move the serialization of username and password configuration into the base class.

PR #153, opened 2017-12-18
Use TripleCollection to look for nodes 

tpluscode:
This is a bigger one.

The current implementation of Graph#GetUriNode and its companions are pretty slow because

they do a Distinct underneath on the subject and object nodes
more importantly, they bypass indexes by enumeration everything
By using Triples.WithSubject(node) and Triples.WithObject(node) the performance boost observed in indexed Graph is multiple orders of magnitude large 😁


kal:
Great fix, thanks @tpluscode !

PR #157, opened 2018-01-10
Add RDF Handler to strip explicit xsd:string datatype from object literals

langsamu:
Based on an answer by @rvesse to [this StackOverflow question](https://stackoverflow.com/questions/40027774/how-to-compare-simple-and-typed-literals-in-dotnetrdf).


kal:
Thanks for the PR!

PR #165, opened 2018-01-22
Relax dateTime coercion in JSON-LD parser

### LINKED ISSUE #164: Unexpected date time handling in JSON-LD parser
JsonLdParser produces some weird results when reading dateTime values:
- Literal values that don't say they're dateTimes are parsed as dateTimes.
- The time zone of the machine running the code has an effect on the parsed value (for the below my time zone was UTC).
- Negative time zone designators are converted to positive ones.
- Values without time zone designators increase in resolution.

Input RDF literal value | Data type | Output graph node literal value
-- | -- | --
2000-01-01T00:00:00 | ex:someDatatype | 2000-01-01T00:00:00.0000000
2000-01-01T00:00:00 | xsd:dateTime | 2000-01-01T00:00:00.0000000
2000-01-01T00:00:00+01:00 | ex:someDatatype | 1999-12-31T23:00:00+00:00
2000-01-01T00:00:00+01:00 | xsd:dateTime | 1999-12-31T23:00:00+00:00
2000-01-01T00:00:00-00:00 | ex:someDatatype | 2000-01-01T00:00:00+00:00
2000-01-01T00:00:00-00:00 | xsd:dateTime | 2000-01-01T00:00:00+00:00
2000-01-01T00:00:00-01:00 | ex:someDatatype | 2000-01-01T01:00:00+00:00
2000-01-01T00:00:00-01:00 | xsd:dateTime | 2000-01-01T01:00:00+00:00

[JsonLdParser.Load](https://github.com/dotnetrdf/dotnetrdf/blob/master/Libraries/dotNetRDF/Parsing/JsonLdParser.cs#L108) uses JsonTextReader with default options, which means we're trying real hard to coerce values.

langsamu:
Fixes #164.

### DIFF
--- a/Testing/unittest/JsonLd/JsonLdParserTests.cs
+++ b/Testing/unittest/JsonLd/JsonLdParserTests.cs
@@ -68,7 +68,7 @@ namespace VDS.RDF.JsonLd
         }
 
         [Theory]
-        [MemberData(nameof(DateTimeValues), DisableDiscoveryEnumeration = false)]
+        [MemberData(nameof(DateTimeValues))]
         public void DateTimeHandlingDiff(string dateTimeValue, string datatype)
         {
             using (var original = new TripleStore())
@@ -85,7 +85,7 @@ namespace VDS.RDF.JsonLd
         }
 
         [Theory]
-        [MemberData(nameof(DateTimeValues), DisableDiscoveryEnumeration = false)]
+        [MemberData(nameof(DateTimeValues))]
         public void DateTimeHandling(string dateTimeValue, string datatype)
         {
             using (var store = new TripleStore())
--- a/Testing/unittest/JsonLd/JsonLdParserTests.cs
+++ b/Testing/unittest/JsonLd/JsonLdParserTests.cs
@@ -1,6 +1,8 @@
 ﻿using FluentAssertions;
+using System;
 using System.Collections.Generic;
 using System.Linq;
+using VDS.RDF.Nodes;
 using VDS.RDF.Parsing;
 using VDS.RDF.Writing;
 using Xunit;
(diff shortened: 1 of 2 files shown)


Testing/unittest/JsonLd/JsonLdParserTests.cs
```diff
@@ -44,5 +45,66 @@ public void TestIssue122()
...
+                    new object[] { "2000-01-01T00:00:00+00:00", "http://example.com/datatype" },
+                    new object[] { "2000-01-01T00:00:00-01:00", "http://example.com/datatype" },
+                    new object[] { "2000-01-01T00:00:00+01:00", "http://example.com/datatype" },
+                };
+            }
+        }
+
+        [Theory]
+        [MemberData(nameof(DateTimeValues), DisableDiscoveryEnumeration = false)]
+        public void DateTimeHandlingDiff(string dateTimeValue, string datatype)
```
tpluscode: Please rename the test method to something more descriptive.

I know, it's not nearly perfect throughout dotNetRDF but we must start somewhere ;)

tpluscode:
Hey @langsamu. Thanks for the fix!

I see in the parsing test that the literal is compared as strings. Maybe for literals typed as actual `xsd:date`/`xsd:datetime` it should be a strongly typed [`DateTimeNode`](https://github.com/dotnetrdf/dotnetrdf/blob/master/Libraries/dotNetRDF/Nodes/DateTimeNode.cs), @kal?

kal:
It looks like the build for the last commit failed because of Chocolatey having some server issues. I've kicked off a rebuild of that.

@tpluscode I think it would be good to have the comparisons use DateTimeNode. @langsamu do you think you could take a look at changing the code that compares using the ILiteralNode interface to use DateTimeNode instead?

Cheers

Kal

langsamu:
8010c9d4ddaea26f154b4f205ad80edbbebef6d0 adds a test that uses [AsValuedNode().AsDateTime()](https://github.com/dotnetrdf/dotnetrdf/commit/8010c9d4ddaea26f154b4f205ad80edbbebef6d0#diff-127d71da23f658dad561d79acedbccadR131).
This succeeds with or without the [DateParseHandling change](https://github.com/dotnetrdf/dotnetrdf/pull/165/commits/87e7a447c3482a681e898b00980ad0ce871790aa#diff-2006619a765ceb0da4f53d6a57798b78R108) in 87e7a447c3482a681e898b00980ad0ce871790aa.

@kal, @tpluscode, is this what you meant?

I've added a strongly typed test instead of modifying the [previous one](https://github.com/dotnetrdf/dotnetrdf/pull/165/commits/ab3eec0d6ad6cc58b984ed6431cfb37c70f91732#diff-127d71da23f658dad561d79acedbccadR105) because I think it's reasonable to expect that `ILiteralNode.Value` is the same as the value of the literal object in the input RDF string.

Similarly I think it's reasonable to [expect](https://github.com/dotnetrdf/dotnetrdf/pull/165/commits/ab3eec0d6ad6cc58b984ed6431cfb37c70f91732#diff-127d71da23f658dad561d79acedbccadR82) that if
N-Triples -> graph1 -> JsonLdWriter -> JSON-LD -> JsonLdParser -> graph2
then `graph1.Difference(graph2).AreEqual`, where strongly typed literal nodes don't play a part (that I know).

kal:
I've had to revert the merge of this PR. The code works fine on my local development environment and the tests pass for all three target frameworks. However on the Appveyor build, the literal round trip tests are failing for some reason that I can't figure out.

langsamu:
@kal: I think this is because your local dev env timezone is different from Appveyor's (as is mine).
35d5fdd changes the strongly typed test so it normalizes to UTC, in line with [ValuedNodeExtensions](https://github.com/dotnetrdf/dotnetrdf/blob/master/Libraries/dotNetRDF/Nodes/ValuedNodeExtensions.cs#L100).
All tests pass on #182.

PR #182, opened 2018-04-22
Relax dateTime coercion in JSON-LD parser

### LINKED ISSUE #164: Unexpected date time handling in JSON-LD parser

langsamu:
Fixes #164, resubmits #165 to [troubleshoot failed tests on Appveyor](https://github.com/dotnetrdf/dotnetrdf/pull/165#issuecomment-383311365).

### DIFF
--- a/Testing/unittest/JsonLd/JsonLdParserTests.cs
+++ b/Testing/unittest/JsonLd/JsonLdParserTests.cs
@@ -1,6 +1,7 @@
 ﻿using FluentAssertions;
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Linq;
 using VDS.RDF.Nodes;
 using VDS.RDF.Parsing;
@@ -111,7 +112,7 @@ namespace VDS.RDF.JsonLd
         [MemberData(nameof(DateTimeValues))]
         public void RoundtripsDatetimeTypedLiterals(string dateTimeValue, string datatype)
         {
-            var isValidDateTime = DateTime.TryParse(dateTimeValue, out DateTime parsedDateTime);
+            var isValidDateTime = DateTime.TryParse(dateTimeValue, null, DateTimeStyles.AdjustToUniversal, out DateTime parsedDateTime);
             var isDateTimeDatatype = datatype == XmlSpecsHelper.XmlSchemaDataTypeDateTime;
 
             if (isValidDateTime && isDateTimeDatatype)


tpluscode:
So only change was needed in the test and not the library itself?

langsamu:
Yes, for [this specific problem](https://github.com/dotnetrdf/dotnetrdf/pull/165#issuecomment-383311365) the [change was to the test only](https://github.com/dotnetrdf/dotnetrdf/pull/182/commits/35d5fdd511f4c4d3ccd6d14d8f2f8d2728e77d0e#diff-127d71da23f658dad561d79acedbccadR115) (because I'd ignored timezones).

However I think the issue and PR are valid regardless.
The test being fixed here is the strongly typed one [you asked for](https://github.com/dotnetrdf/dotnetrdf/pull/165#issuecomment-359528983), which shows how the issue doesn't apply when you use ValuedNodes.
But when [not relying on ValuedNodes](https://github.com/dotnetrdf/dotnetrdf/pull/182/files#diff-127d71da23f658dad561d79acedbccadR90) the issue does apply without the fix.

PR #184, opened 2018-05-03
GraphML IStoreWriter

langsamu:


Libraries/dotNetRDF/Writing/GraphMLWriter.cs
```diff
@@ -0,0 +1,282 @@
...
+            writer.WriteAttributeString(GraphMLHelper.Key, key);
+            writer.WriteString(value);
+            writer.WriteEndElement();
+        }
+    }
+
+    /// <summary>
+    /// A helper class containing GraphML name and URI constants
+    /// </summary>
+    public static class GraphMLHelper
```
tpluscode: Please put classes in individual files

tpluscode: Besides, I don't particularly like the name but I think we have other similar classes for Xsd, etc, right? @kal maybe we could put `GraphMLHelper` in the same namespace?

Testing/unittest/Writing/GraphMLTests.cs
```diff
@@ -0,0 +1,145 @@
...
+
+using System;
+using System.Linq;
+using System.Xml;
+using System.Xml.Linq;
+using Xunit;
+
+namespace VDS.RDF.Writing
+{
+    public class GraphMLTests : IClassFixture<GraphMLFixture>
```
tpluscode: Oh wow, I never knew about `IClassFixture`. Is this something new in more recent xUnit?

Testing/unittest/Writing/GraphMLTests.cs
```diff
@@ -0,0 +1,145 @@
...
+            reader.Settings.Schemas.Add(GraphMLHelper.NS, GraphMLHelper.XsdUri);
+            reader.Settings.ValidationType = ValidationType.Schema;
+            reader.Settings.ValidationEventHandler += (sender, e) => throw e.Exception;
+
+            while (reader.Read()) { }
+        }
+#endif
+    }
+
+    public class GraphMLFixture
```
tpluscode: Please move to separate file

Libraries/dotNetRDF/Writing/GraphMLWriter.cs
```diff
@@ -0,0 +1,282 @@
...
+
+            foreach (var graph in store.Graphs)
+            {
+                GraphMLWriter.WriteGraph(writer, graph);
+            }
+
+            writer.WriteEndElement();
+        }
+
+        private static void WriteGraph(XmlWriter writer, IGraph graph)
```
tpluscode: I'm worried that this can be not too performant with multiple loops over the graph. Have you tried it on larger datasets?

Libraries/dotNetRDF/Writing/GraphMLWriter.cs
kal: @tpluscode, @langsamu 

+1 for separate files for each class. I know that this is not the case with the rest of the code-base and I think it is probably a matter of personal style, but I find it difficult to work with code where multiple classes are combined in a single source file.

The xxx**SpecsHelper** name is quite a common pattern in the code base, so can we please rename this class to `GraphMLSpecsHelper`? Many of those `xxxSpecsHelper` classes sit in the Parsing namespace as they are used by both parsers and writers. In this case we won't have a GraphML parser, so I'm OK with the GraphMLSpecsHelper going in the Writing namespace.

Libraries/dotNetRDF/Writing/GraphMLWriter.cs
```diff
@@ -0,0 +1,282 @@
...
+            if (graph.BaseUri != null)
+            {
+                writer.WriteAttributeString(GraphMLHelper.Id, graph.BaseUri.AbsoluteUri);
+            }
+
+            // RDF is always a directed graph
+            writer.WriteAttributeString(GraphMLHelper.Edgedefault, GraphMLHelper.Directed);
+
+            // The next two iterations are disjoint
+            foreach (var literalNode in graph.Nodes.Where(n => n.NodeType != NodeType.Literal))
```
kal: The variable here is called `literalNode`, but the test is for non-literal nodes.

Libraries/dotNetRDF/Writing/GraphMLWriter.cs
```diff
@@ -0,0 +1,282 @@
...
+            foreach (var triple in graph.Triples.Where(t => t.Object.NodeType == NodeType.Literal))
+            {
+                GraphMLWriter.WriteNode(writer, triple.GetHashCode().ToString(), triple.Object.ToString());
+            }
+
+            foreach (var triple in graph.Triples)
+            {
+                GraphMLWriter.WriteEdge(writer, triple);
+            }
+
```
kal: Would it be possible to merge the above two loops into a single one that writes the literal node if the object is a literal and then writes the edge? AFAIK GraphML doesn't care about the ordering of nodes/edges in the file.

Libraries/dotNetRDF/Writing/GraphMLWriter.cs
tpluscode: Fair enough, let's keep it here.

Testing/unittest/Writing/GraphMLTests.cs
langsamu: Moved.

Libraries/dotNetRDF/Writing/GraphMLWriter.cs
langsamu: Doesn't matter now the three iterations are one.

langsamu:
Thanks for the feedback.

I've implemented the main functionality in static methods because they use no instance variables. I think [there's a minor performance gain](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1822-mark-members-as-static), but more important to me is the message "this code is stateless".

The XmlWriter could be an instance variable instead of a parameter, but then I'd have to worry about multiple invocations of the interface methods.

PR #190, opened 2018-05-11
Decouple HtmlWriter from SubjectNodes

### LINKED ISSUE #189: HtmlWriter breaks on retraction
h/t @christopheralcock 

When triples are removed from a graph prior to saving it with an `HtmlWriter`, the generated HTML is broken, with extra `tr`s for retracted statements and `td rowspan="0"` for removed subject nodes.

This is becuase [HtmlWriter relies](https://github.com/dotnetrdf/dotnetrdf/blob/350fb42236d1c048a6503712043dd76e80d7d247/Libraries/dotNetRDF/Writing/HtmlWriter.cs#L190) on `BaseTripleCollection.SubjectNodes`, which is broken for indexed graphs (#188).

screenshot

langsamu:
Fixes #189.

### DIFF
--- a/Libraries/dotNetRDF/Writing/HtmlWriter.cs
+++ b/Libraries/dotNetRDF/Writing/HtmlWriter.cs
@@ -95,7 +95,7 @@ namespace VDS.RDF.Writing
         /// <param name="output">Stream to save to</param>
         /// <param name="leaveOpen">Whether to leave <paramref name="output"/> open after writing the graph</param>
         public void Save(IGraph g, TextWriter output, bool leaveOpen)
-        { 
+        {
             try
             {
                 g.NamespaceMap.Import(_defaultNamespaces);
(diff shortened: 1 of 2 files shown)


Libraries/dotNetRDF/Writing/HtmlWriter.cs
```diff
@@ -187,7 +187,9 @@ private void GenerateOutput(HtmlWriterContext context)
             context.HtmlWriter.RenderBeginTag(HtmlTextWriterTag.Tbody);
 
             TripleCollection triplesDone = new TripleCollection();
-            foreach (INode subj in context.Graph.Triples.SubjectNodes)
+            
+            // HACK: Getting subject nodes explicitly while BaseTripleCollection.SubjectNodes is broken for indexed graphs
+            foreach (INode subj in context.Graph.Triples.Select(t => t.Subject).Distinct())
```
kal: I can see why you do this as it is probably the only workaround for the in-memory indexed graph, but I'm concerned that the .Distinct() is not scalable for larger graphs (not so much the in-memory graphs, but ones where the triples might be materialized from some backing store).

I wonder if it might be better to have a look at fixing SubjectNodes instead.

kal:
#192 has addressed this problem.

Issue #204, opened 2018-09-29
TriG is as old as the hills


langsamu:
QNames with local part that starts with a digit cannot be parsed as TriG:
```c#
new TripleStore().LoadFromString(
    "@prefix : <http://example.com/> . { :1 :p :o . }",
    new TriGParser());
```
The [tokeniser](https://github.com/dotnetrdf/dotnetrdf/blob/master/Libraries/dotNetRDF/Parsing/Tokens/TriGTokeniser.cs#L469) uses TurtleSpecHelper, which [defaults](https://github.com/dotnetrdf/dotnetrdf/blob/master/Libraries/dotNetRDF/Parsing/TurtleSpecsHelper.cs#L200) to the [original Turtle syntax](https://www.w3.org/TeamSubmission/turtle/), where [`nameStartChar`](https://www.w3.org/TeamSubmission/turtle/#nameStartChar) can't be a digit.

Latest TriG [does allow digits](https://www.w3.org/TR/trig/#grammar-production-PNAME_LN), just like [Turtle in DNR](https://github.com/dotnetrdf/dotnetrdf/blob/master/Libraries/dotNetRDF/Parsing/TurtleSpecsHelper.cs#L343).

PR #205, opened 2018-09-29
First step to support modern TriG

langsamu:
Addresses #204 by choosing recent Turtle syntax for parsing recent TriG.


kal:
Thanks @langsamu - this is definitely a useful addition to the TriG parser.

PR #214, opened 2018-10-21
VALUES support for query builder

### LINKED ISSUE #211: Building SPARQL with VALUES clause
On the Basic Tutorial section **UserGuide Building SPARQL**, there is no indication of how to create a VALUES clause within a SPARQL. For example, 

`SELECT ?x
 WHERE { VALUES (?x) { (http://www.w3.org/2001/vcard-rdf/3.0#FN) } }`

How can I create a SPARQL with a VALUES clause using the `QueryBuilder`?

tpluscode:
WIP adding `VALUES` to query builder API:

* [x] adding `VALUES` to graph patterns
  * [x] tagged and typed literals
  * [x] URIs
* [x] adding `VALUES` to entire query
* [x] XML doc

fixes #211

### DIFF
--- a/Libraries/dotNetRDF/Query/Builder/Expressions/LiteralExpressionExtensions.cs
+++ b/Libraries/dotNetRDF/Query/Builder/Expressions/LiteralExpressionExtensions.cs
@@ -33,58 +33,37 @@ namespace VDS.RDF.Query.Builder.Expressions
     {
         private static readonly NodeFactory NodeFactory = new NodeFactory();
 
-        private static ILiteralNode ToLiteral<T>(T numericValue)
+        internal static ILiteralNode ToLiteral(object value)
         {
-            if (typeof(T) == typeof(int))
+            switch (value)
             {
-                return ((int)(object)numericValue).ToLiteral(NodeFactory);
+                case int i:
+                    return i.ToLiteral(NodeFactory);
+                case decimal d:
+                    return d.ToLiteral(NodeFactory);
+                case short s:
+                    return s.ToLiteral(NodeFactory);
+                case long l:
+                    return l.ToLiteral(NodeFactory);
+                case float f:
+                    return f.ToLiteral(NodeFactory);
+                case double d:
+                    return d.ToLiteral(NodeFactory);
+                case byte b:
+                    return b.ToLiteral(NodeFactory);
+                case sbyte s:
+                    return s.ToLiteral(NodeFactory);
+                case string s:
+                    return new LiteralNode(null, s);
+                case DateTime d:
+                    return d.ToLiteral(NodeFactory);
+                case TimeSpan t:
+                    return t.ToLiteral(NodeFactory);
+                case bool b:
+                    return b.ToLiteral(NodeFactory);
+                default:
+                    throw new ArgumentException(string.Format("Unsupported type for literal node: {0}", value.GetType()));
             }
-            if (typeof(T) == typeof(decimal))
-            {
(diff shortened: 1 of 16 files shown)


Libraries/dotNetRDF/Query/Builder/QueryBuilder.cs
```diff
@@ -377,10 +384,7 @@ private void BuildRootGraphPattern(SparqlQuery query)
         {
             var rootGraphPattern = RootGraphPatternBuilder.BuildGraphPattern(Prefixes);
 
-            if (!rootGraphPattern.IsEmpty)
-            {
-                query.RootGraphPattern = rootGraphPattern;
-            }
+            query.RootGraphPattern = rootGraphPattern;
```
tpluscode: This one change seems unrelated but I found that `VALUES` to not switch `IsEmpty`. 

Does is make any difference if the root in null or not?

Libraries/dotNetRDF/Query/Builder/Expressions/LiteralExpressionExtensions.cs
```diff
@@ -33,58 +33,37 @@ internal static class LiteralExpressionExtensions
     {
         private static readonly NodeFactory NodeFactory = new NodeFactory();
 
-        private static ILiteralNode ToLiteral<T>(T numericValue)
+        internal static ILiteralNode ToLiteral(object value)
         {
-            if (typeof(T) == typeof(int))
+            switch (value)
```
tpluscode: Turned those ifs into C# 7.1 pattern matching of generic type argument. Do you have any objections? I understand that it only affect build-time requirements?

Libraries/dotNetRDF/Query/Builder/IQueryBuilder.cs
```diff
@@ -41,8 +41,13 @@ public interface IQueryBuilder
         /// <summary>
         /// Gets the builder associated with the root graph pattern.
         /// </summary>
+        [Obsolete("Please use Root property")]
         GraphPatternBuilder RootGraphPatternBuilder { get; }
         /// <summary>
+        /// Gets the builder associated with the root graph pattern.
+        /// </summary>
+        IGraphPatternBuilder Root { get; }
```
tpluscode: Added a new alias and made the long property obsolete bc it simply seems overly verbose

Libraries/dotNetRDF/Query/Builder/Expressions/LiteralExpressionExtensions.cs
kal: No objection from me - I tend to update C# syntax as I go along in any case.

Libraries/dotNetRDF/Query/Builder/QueryBuilder.cs
kal: I think if the root graph pattern is null / empty then you should just end up with the values in the result set. Not 100% sure, but I don't think this is a problem ;-)

Libraries/dotNetRDF/Query/Builder/IQueryBuilder.cs
kal: 👍

kal:
This is looking really good!

tpluscode:
Added XML comments. Please check my latest commit and we're ready to merge

kal:
Thanks @tpluscode!

assumption:
Thank you @tpluscode for this!

PR #221, opened 2018-11-09
Increase performance when creating OntologyClass and OntologyProperty…

DFihnn:
… by using a Hashset instead of doing contains checks on a List to avoid duplicates.


kal:
This looks good - I can't see any dependency on the ordered nature of the list so I'm happy to merge this PR.

Thanks @DFihnn !

Issue #224, opened 2019-01-16
Support Stardog transactions enabled for reasoning queries


DarthStem:
Stardog only supports reasoning queries into a transaction session if the transaction has been opened with reasoning on.  
See https://community.stardog.com/t/stardog-5-query-with-reasoning-returning-0-results-when-run-inside-transaction/447/6

This support may have been missed during the PR which addressed Issue #108 but that's not completely clear based on when Stardog altered their behavior in v5.3.0.

PR #226, opened 2019-02-05
StarDog reasoning query support in a open transaction

DarthStem:
Changes needed to address Issue #224

### DIFF
--- a/Libraries/dotNetRDF/Storage/StardogConnector.cs
+++ b/Libraries/dotNetRDF/Storage/StardogConnector.cs
@@ -457,7 +457,7 @@ namespace VDS.RDF.Storage
                     if (reasoning)
                     {
                         postData.Append("reasoning=");
-                        postData.Append(reasoning.ToString()+"&");
+                        postData.Append(reasoning.ToString() + "&");
                     }
                     postData.Append("query=");
                     postData.Append(HttpUtility.UrlEncode(sparqlQuery));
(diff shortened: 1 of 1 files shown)


Libraries/dotNetRDF/Storage/StardogConnector.cs
```diff
@@ -1932,10 +1932,10 @@ public virtual void Query(string query, AsyncStorageCallback callback, object st
                 }
             }
 
-            requestUri += GetReasoningParameter();
+            if (Reasoning != StardogReasoningMode.DatabaseControlled) requestUri += GetReasoningParameter();
```
kal: This isn't really needed as GetReasoningParameter returns an empty string if ReasoningMode is set to DatabaseControlled. I'd rather leave the logic of setting the request parameter encapsulated in that method than add an exception here.

Libraries/dotNetRDF/Storage/StardogConnector.cs
```diff
@@ -2021,19 +2021,23 @@ protected virtual string GetReasoningParameter()
         /// <summary>
         /// Start a transaction
         /// </summary>
+        /// <param name="enableReasoning">Opens the transaction with reasoning enabled.</param>
```
kal: As this connector supports older versions of Stardog, I think it would be worth adding a note here that reasoning inside a transaction is only supported in Stardog v5.3.0 onwards.

Libraries/dotNetRDF/Storage/StardogConnector.cs
DarthStem: Reverted back to the original logic

PR #228, opened 2019-02-13
Dynamic graphs and nodes

langsamu:
Inspired by @mattrayner's [GROM](https://github.com/ukparliament/grom), which is used extensively in @ukparliament's [new website](https://beta.parliament.uk/). Used in [this ASP.NET Core SKOS viewer](https://github.com/ukparliament/VocabularyBrowser) for [Parliament's Controlled Vocabulary](https://api.parliament.uk/vocabulary/browser)

---
# Dynamic graphs and nodes
This feature adds read/write dictionary and dynamic capabilities to `IGraph` and `INode` by implementing `IDictionary<>` and `IDynamicMetaObjectProvider` on `WrapperGraph` and `WrapperNode`.

## Examples
Tests in the Example class ([1](https://github.com/langsamu/dotnetrdf/blob/4b5dcc0a70c5480e99c0876604de4b30bc4787e0/Testing/unittest/Dynamic/Examples.cs), [2](https://github.com/langsamu/dotnetrdf/blob/4b5dcc0a70c5480e99c0876604de4b30bc4787e0/Testing/unittest/Dynamic/Examples.NetFull.cs)) give an overview of feature details.

---
**I'd appreciate feedback and help making this PR better.**

### DIFF
--- a/Libraries/dotNetRDF/Core/WrapperNode.NetFull.cs
+++ b/Libraries/dotNetRDF/Core/WrapperNode.NetFull.cs
@@ -1,0 +1,61 @@
+﻿/*
+// <copyright>
+// dotNetRDF is free and open source software licensed under the MIT License
+// -------------------------------------------------------------------------
+// 
+// Copyright (c) 2009-2017 dotNetRDF Project (http://dotnetrdf.org/)
+// 
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is furnished
+// to do so, subject to the following conditions:
+// 
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+// 
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR 
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, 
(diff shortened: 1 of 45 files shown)


kal:
This looks great @langsamu  - I think it is the kind of thing that definitely helps the use of RDF in a .NET environment and I'd be really happy to accept this into the DNR code-base. Its a lot to review here so I'm going to have a bit of a play around with the code today and if I have any other feedback I'll add it here. 

Two initial comments though:

Can you add a page or two to the wiki that documents this stuff. I'd like to link that in then to the sequence on Getting Started, because I think for a lot of developers using the dynamic API might be preferable to using the low-level API if they can get their head around the property name to URI mappings.

I can see how you could use this API over the result of a SPARQL CONSTRUCT / DESCRIBE query, which I think is a massive benefit. I wonder if adding a bit more code to enable a tabular result set from a SELECT query to be mapped to an enumeration of dynamic objects might also be something worth considering so that there isn't a sudden disconnect from the dyamic API when doing a SPARQL query.

langsamu:
Thanks for the comments @kal.

1. I've implemented dynamic behaviour for SELECTs in 753d3b1. Thanks for the idea.
2. I've started working on [documentation](https://github.com/langsamu/dotnetrdf/wiki/UserGuide-Dynamic-API) for the Dynamic API. Covers a theoretical overview and explains predicate mappings. Still needs explanation of assignment conversions and a good structure.

kal:
Thanks @langsamu - I'm happy to merge the PR as it stands, but the docs probably need a bit more work. I've started to add a bit more "how to" documentation to the start before diving into the theory as I think it would be good to just provide some quick start documentation for the API. What I've added so far is based on the unit tests and examples, but if you could give it a quick sense-check to make sure I haven't misrepresented anything that would be great!

Still need to add documentation about Contains and ContainsKey and about accessing SPARQL result sets.

kal:
BTW - I've merged the docs to the DNR wiki too.

PR #234, opened 2019-03-24
Fix URI string for language-tagged string

### LINKED ISSUE #233: Wrong URI for language-tagged string
https://github.com/dotnetrdf/dotnetrdf/blob/e27fce94772d574817f4c7a510aed2b58803cb0c/Libraries/dotNetRDF/Parsing/RDFSpecsHelper.cs#L80

Should be `langString`, not `rdfLangString`.
Looks like a typo or copy/paste artefact to me.

langsamu:
Fixes #233.


kal:
Thanks @langsamu !

PR #239, opened 2019-04-02
Changed sorting for TurtleWriter.GenerateOutput(TurtleWriterContext c…

MischaVreeburg:
…ontext)

Altered the List<triple>.Sort to use a layered dictionary (dictionary(subjectnode, dictionary( predicatenode, dictionary (objectnode, list<triple>))))
 and sorting the keys for each dictionary using the FastNodeComparer. Then use a list sort on each of the list<triples>, which is probable superfluous but if there are more than one it should sort, using FullTripleComparer(new FastNodeComparer()).

Because each of the sets are much smaller the sorting will be much faster.

### DIFF
--- a/Libraries/dotNetRDF/Writing/TurtleWriter.cs
+++ b/Libraries/dotNetRDF/Writing/TurtleWriter.cs
@@ -2,21 +2,21 @@
 // <copyright>
 // dotNetRDF is free and open source software licensed under the MIT License
 // -------------------------------------------------------------------------
-// 
+//
 // Copyright (c) 2009-2017 dotNetRDF Project (http://dotnetrdf.org/)
-// 
+//
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
 // in the Software without restriction, including without limitation the rights
 // to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
 // copies of the Software, and to permit persons to whom the Software is furnished
 // to do so, subject to the following conditions:
-// 
+//
 // The above copyright notice and this permission notice shall be included in all
 // copies or substantial portions of the Software.
-// 
-// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR 
-// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, 
+//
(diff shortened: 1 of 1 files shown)


Libraries/dotNetRDF/Writing/TurtleWriter.cs
```diff
@@ -182,21 +167,60 @@ private void GenerateOutput(TurtleWriterContext context)
...
+                int capacity = ts.Count;
+                Dictionary<INode, Dictionary<INode, Dictionary<INode, List<Triple>>>> sortHelperDictionary = new Dictionary<INode, Dictionary<INode, Dictionary<INode, List<Triple>>>>(capacity);
+                // Fill dictionary
+                foreach (Triple triple in ts)
+                {
+                    if (!sortHelperDictionary.ContainsKey(triple.Subject)) { sortHelperDictionary.Add(triple.Subject, new Dictionary<INode, Dictionary<INode, List<Triple>>>()); }
+
+                    if (!sortHelperDictionary[triple.Subject].ContainsKey(triple.Predicate)) { sortHelperDictionary[triple.Subject].Add(triple.Predicate, new Dictionary<INode, List<Triple>>()); }
+
+                    if (!sortHelperDictionary[triple.Subject][triple.Predicate].ContainsKey(triple.Object)) { sortHelperDictionary[triple.Subject][triple.Predicate].Add(triple.Object, new List<Triple>()); }
```
kal: I think this level of the dictionary is not really needed. Objects don't have to be output in sort order, we just need the ordering to allow grouping by subject and predicate.

Libraries/dotNetRDF/Writing/TurtleWriter.cs
```diff
@@ -182,21 +167,60 @@ private void GenerateOutput(TurtleWriterContext context)
...
+                        foreach (var objectKey in objectKeys)
+                        {
+                            var triplesSubset = sortHelperDictionary[subjectKey][predicatekey][objectKey];
+                            if (triplesSubset.Count > 1)
+                            {
+                                triplesSubset.Sort(new FullTripleComparer(new FastNodeComparer()));
+                            }
+
+                            ts.AddRange(triplesSubset);
+                        }
```
kal: If we don't have the object level of the sortHelperDictionary then this inner loop isn't needed so you could just add all the triples for a given subject+predicate pair.

MischaVreeburg: I added this level to get the same sorting as with List.Sort. If you say we don't need it, I'll remove it.

MischaVreeburg:
Well, we all need our holidays. Thanks for the review.

PR #240, opened 2019-04-02
Feature/triple collection performance

MischaVreeburg:
Changed the default selection options on BaseTripleCollection for the triples containing nodes from using Linq where to foreach loop with yield return. This has a better performance with large sets.

Also removed the double enumeration when selecting for two nodes.

Added 6 multidictionarys to the TripleCollection. 
Triples are also stored in these and can be accessed by their nodes.
Three dictionaries for single node access (subject, predicate, object)
and three dictionaries for multi node access (subject-predicate, predicate-object, subject-object).

Changed Node getters for TripleCollection to return the keyCollections from the first set of dictionaries instead of running two linq queries (massive performance gain)
Added overrides for default queries for triple list containing specific nodes. Now returns dictionary value for node key(s) => HashSet<Triple>  (massive performance gain)

Added correctly adding and removing of triples for extra dictionaries (slight performance hit)

### DIFF
--- a/Libraries/dotNetRDF/Core/BaseTripleCollection.cs
+++ b/Libraries/dotNetRDF/Core/BaseTripleCollection.cs
@@ -2,21 +2,21 @@
 // <copyright>
 // dotNetRDF is free and open source software licensed under the MIT License
 // -------------------------------------------------------------------------
-// 
+//
 // Copyright (c) 2009-2017 dotNetRDF Project (http://dotnetrdf.org/)
-// 
+//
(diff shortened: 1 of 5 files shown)


kal:
Hi @MischaVreeburg 

Thanks for this, but to me it looks like this PR changes TripleCollection to be the same (or pretty much the same) as TreeIndexedTripleCollection, which is the default triple collection used by the in-memory graph implementation.

When you are comparing performance are you comparing with the default graph implementation? Can you provide some numbers or some test case that shows the performance hit using TreeIndexedTripleCollection.

Cheers

Kal

PR #268, opened 2019-10-02
Add ObjectUri. Closes #266

### LINKED ISSUE #266: How to create this Query: ?item wdt:P31 wd:Q146.
Hi community!

I have the following:
``` csharp
.Where(
  (triplePatternBuilder) =>
  {
    triplePatternBuilder
      .Subject(cat)
      .PredicateUri("wdt:P31")
      .Object("wd:Q146");
  })
```
Which returns `{ ?item wdt:P31 ?wd:Q146 . } `
I also tried `.Object(new Uri("wd:Q146"))`
`{ ?item wdt:P31 <wd:Q146> . } `

But not of those are what I am looking for. Any clues?

Thanks !

gabrieldelaparra:


Libraries/dotNetRDF/Query/Builder/TriplePatternObjectPart.cs
```diff
@@ -129,5 +129,15 @@ public ITriplePatternBuilder Object(PatternItem objectPattern)
             _triplePatternBuilder.AddPattern(new TriplePattern(_subjectPatternItem, _predicatePatternItem, objectPattern));
             return _triplePatternBuilder;
         }
+
+        /// <summary>
+        /// Depending on the generic parameter type, sets a literal, a QName or a blank node as <see cref="IMatchTriplePattern.Object"/>
+        /// </summary>
+        /// <param name="objectUri">Either a variable name, a literal, a QName or a blank node identifier</param>
+        /// <remarks>A relevant prefix/base URI must be added to <see cref="IQueryBuilder.Prefixes"/> to accept a QName</remarks>
```
tpluscode: Please adjust the XML doc. This method only accepts prefixed names. You might also want to rename the parameter so that it's clear that fully qualified URIs will not work.

gabrieldelaparra:
Hi Kal,

Thanks for your reply.
I thought the same thing, but only after finding out about the custom
derived class.

Cheers,

PR #294, opened 2020-01-05
SHACL processor constraint default result messages

### LINKED ISSUE #293: Adding default result messages for constraints (Shacl)
First of all a big praise to @langsamu  for the development of the shacl validator. After first tests it works very well. 

Unfortunately there are no standardized messages for the constraints of the shacls. Although the specification does not require such messages, but they would be a great help for using the validator. 

[The spec says in 3.6.2.7](https://www.w3.org/TR/shacl/#results-message)
> In cases where a constraint does not have any values for sh:message in the shapes graph the SHACL processor may automatically generate other values for sh:resultMessage

Is there already a plan for adding such messages or can you please include this in your roadmap?

langsamu:
Resolves #293.

### DIFF
--- a/Libraries/dotNetRDF/Shacl/Constraint.cs
+++ b/Libraries/dotNetRDF/Shacl/Constraint.cs
@@ -46,6 +46,26 @@ namespace VDS.RDF.Shacl
 
         protected Shape Shape { get; private set; }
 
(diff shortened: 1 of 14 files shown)


kal:
Hi @langsamu. Did you want some more input / help on this?

Issue #278, opened 2019-12-02
Stardog Connector constructor/Stardog server GetStore does not validate store exists


joephayes:
The StardogConnector constructor does not validate that store passed into it actually exists. So this code will not throw an error:

```
StardogConnector stardogConnector = new StardogConnector("http://localhost:5820", "i_dont_exist", "admin", "admin");
```

This code is called by the StardogServer GetStore method, so this code won't throw an error either:

``` 
StardogServer stardogServer = new StardogServer("http://localhost:5820", 
"admin", "admin");
IStorageProvider store = stardogServer.GetStore("i_dont_exist");
```

Not sure if this behavior is standardized at all but I think both cases should validate the store exists (or the `IsReady` property should at least return `FALSE`) because anything the user tries to do with the connector or store will fail.

Happy to contribute a PR...

kal:
Thanks for the report!

That behaviour is actually consistent with the other connectors. All of the constructors are pretty light-weight and don't involve a roundtrip to the server. Whether that is a desireable behaviour or not is another question.

I think that the suggestion that `IsReady` should return `False` if a connection cannot be made to the server is a good one. How that is implemented will have to depend on the store being connected to because SPARQL Protocol provides no standard endpoint for checking the validity of an endpoint.

So I think a good PR would be one that implements some proper endpoint check when IsReady is called for the first time and then caches that value.

PR #299, opened 2020-02-04
IsReady implementation for StardogConnector

joephayes:
Fixes issue #278 

The IsReady property of the StardogConnector will now check if the specified database is actually included in the list of stores returned by a call to the Server's `ListStores` method


kal:
Ignoring codecov report as the build server doesn't run Stardog so can't run the unit test that you added. 
PR merged. Thanks @joephayes !

PR #304, opened 2020-04-09
Check for null before pre-binding $shapesGraph

### LINKED ISSUE #303: SHACL SPARQL fails $shapesGraph outside triplestore
The SHACL processor [uses](https://github.com/dotnetrdf/dotnetrdf/blob/61d0469807f8bc6c174993acd18573e5ad521c67/Libraries/dotNetRDF/Shacl/Constraints/Sparql.cs#L97) `INode.GraphUri` to pre-bind the `shapesGraph` variable for SPARQL constraints, however this is `null` for graphs not loaded from a URI, causing a null dereference.

To reproduce:
```csharp
var shapesGraph = new Graph();
shapesGraph.LoadFromString(@"
@prefix sh: <http://www.w3.org/ns/shacl#> .

[
    sh:targetNode <http://example.com/s> ;
    sh:sparql [ sh:select 'SELECT * { ?this <http://example.com/p> ?o . }' ] ;
] .
");

var dataGraph = new Graph();
dataGraph.LoadFromString(@"<http://example.com/s> <http://example.com/p> <http://example.com/o> .");

new ShapesGraph(shapesGraph).Validate(dataGraph); // throws NullReferenceException
```

While the following succeeds:
```csharp
var shapesGraph = new Graph();
shapesGraph.LoadFromUri(uri); // assume uri serves same RDF as shapes graph above

var dataGraph = new Graph();
dataGraph.LoadFromString("<http://example.com/s> <http://example.com/p> <http://example.com/o> .");

new ShapesGraph(shapesGraph).Validate(dataGraph); // succeeds
```

The spec [says](https://www.w3.org/TR/shacl/#x5.3.1-pre-bound-variables-in-sparql-constraints-($this,-$shapesgraph,-$currentshape)) this variable is optional:
> If the shapes graph is a named graph in the same dataset as the data graph then it is the IRI of the shapes graph in the dataset.

The processor shouldn't fail if the shapes graph is not a named graph.

langsamu:
Fixes #303.

### DIFF
--- a/Libraries/dotNetRDF/Shacl/Constraints/Sparql.cs
+++ b/Libraries/dotNetRDF/Shacl/Constraints/Sparql.cs
@@ -94,7 +94,11 @@ namespace VDS.RDF.Shacl.Constraints
 
             BindFocusNode(query.RootGraphPattern, focusNode);
             query.RootGraphPattern.TriplePatterns.Insert(0, new BindPattern("currentShape", new ConstantTerm(Shape)));
-            query.RootGraphPattern.TriplePatterns.Insert(0, new BindPattern("shapesGraph", new ConstantTerm(Shape.Graph.CreateUriNode(Shape.GraphUri))));
+
+            if (Shape.GraphUri is object)
+            {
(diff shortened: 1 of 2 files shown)


kal:
Thanks @langsamu !

Issue #305, opened 2020-04-21
Leviathan processor gives different results for different Windows regional formats


uyha:
```c#
const string update = @"
DELETE { ?s ?p ?o. }
INSERT { ?s ?p ?a. }
WHERE  { ?s ?p ?o. BIND(?o / 2 AS ?a) }
";
const string query = @"SELECT * WHERE { ?s ?p ?o.}";

var graph = new Graph();
graph.Assert(graph.CreateBlankNode(), graph.CreateBlankNode(), 0.5.ToLiteral(graph));

var store = new TripleStore();
store.Add(graph);

var updateProcessor = new LeviathanUpdateProcessor(store);
var updateParser = new SparqlUpdateParser();
updateParser.ParseFromString(update).Process(updateProcessor);

var queryProcessor = new LeviathanQueryProcessor(store);
var queryParser = new SparqlQueryParser();
var result = queryParser.ParseFromString(query).Process(queryProcessor) as SparqlResultSet;
var @object = result.First()["o"].AsValuedNode().AsDouble();
//@object is 25 when using German (Germany) regional format
//@object is 0.25 when using English (United Kingdom) regional format
```

kal:
Thanks for the bug report. I've reproduced the error and it does look like it is an issue with the way that the literal node is formatted when it is inserted in the update.

The issue also partially also shows up for a simple query such as: `SELECT ?a WHERE { BIND (0.25 as ?a) }`. In this case the double value of the result is correct, but the string value is returned in the regional format (0,25) rather than as a SPARQL double literal string (0.25). My guess is that on update the regionally formatted string is inserted into the triple store using the regional format, but then when you query, that string value is converted to a double using the invariant format and so you get the value being interpreted as 25 rather than 0.25. I'll have to dig a bit to see where the regional formatting is being applied and why that is, but I can't think of a good reason for it to be behaving in the way it does.

Issue #306, opened 2020-04-21
Accessing UpdateExecutionTime of SparqlUpdateCommandSet can throw an exception


ghost:
Hi, we are currently writing an REST API using ASP.NET and dotnetrdf. We implemented an input formatter that reads SPARQL update requests and emits a `SparqlUpdateCommandSet` which can then be used directly as a parameter of a controller method. Unfortunately, something in the ASP.NET framework inspects values emitted by input formatters and reads the `UpdateExecutionTime` field, which throws an exception because the command set has not been executed yet and results in `null` being passed to the controller.

As the field is already a `Nullable` field, would it not make more sense to return `null` if the command set has not been executed? Or maybe return `TimeSpan.Zero`?

We worked around this problem by manually setting `UpdateExecutionTime` to `TimeSpan.Zero` before returning the command set from the input formatter.

kal:
I think if I were designing the SparqlUpdateCommandSet interface from scratch I would probably go with returning null and maybe adding another "HasExecuted" boolean flag to the class rather than throw an exception in the property getter. 

However, my worry with implementing this now is that it changes the behaviour of the class. It is definitely a thing to scope in to a 3.0 release, but it wouldn't be a backward compatible change that we could make in a point release. I think that if your workaround is solid enough for your purposes (and it sounds like it is) then I'd prefer to hold of on making any change in the 2.x releases and instead move this to the 3.0 release.

mrf-ipt:
I think you accidentally referenced the wrong issue in the commit.

kal:
Yes I did! Apologies.

PR #307, opened 2020-04-22
Fix NumericNode constructors

kal:
Fixes the constructors of the classes derived from NumericNode so that the constructor that takes only a value type parameter generates a literal value using the invariant culture.
Addresses #306


kal:
Fixes #305 (not 306)

PR #314, opened 2020-05-15
Decouple default graph node lookups from node implementations

### LINKED ISSUE #313: Node lookups in default graph coupled to node implementations
# Description

The [default graph's implementations of `GetXXXNode`](https://github.com/dotnetrdf/dotnetrdf/blob/master/Libraries/dotNetRDF/Core/Graph.cs#L155-L232) use a common utility method `GetNode<T>` to look up nodes. For example

https://github.com/dotnetrdf/dotnetrdf/blob/8feee66fd8bcec83b9e1f059257324e87ea6b6b7/Libraries/dotNetRDF/Core/Graph.cs#L162-L165

`T` here is inferred to be `UriNode`, the default implementation of `IUriNode`, which is more specific than the caller's return type.

This is a problem whenever using nodes not derived from default implementations (e.g. `WrapperNode`s, which I find myself using more and more).

# To reproduce

```csharp
class MyUriNode : BaseUriNode
{
    public MyUriNode(IGraph g, Uri uri) : base(g, uri) { }
}

class TestClass
{
    void TestMethod()
    {
        var g = new Graph();
        var uri = UriFactory.Create("http://example.com/s");

        g.Assert(
            new MyUriNode(g, uri),
            g.CreateBlankNode(),
            g.CreateBlankNode());

        g.GetUriNode(uri); // System.InvalidCastException : Unable to cast object of type 'MyUriNode' to type 'VDS.RDF.UriNode'.
    }
}
```

# Resolution

Just need to make `T` explicit:

```csharp
return GetNode<IUriNode>(new UriNode(this, uri)); 
```

langsamu:
Fixes #313.

### DIFF
--- a/Libraries/dotNetRDF/Core/Graph.cs
+++ b/Libraries/dotNetRDF/Core/Graph.cs
@@ -39,7 +39,7 @@ namespace VDS.RDF
 #if !NETCORE
     [Serializable,XmlRoot(ElementName="graph")]
 #endif
-    public class Graph 
+    public class Graph
         : BaseGraph
     {
         #region Constructor
@@ -47,7 +47,7 @@ namespace VDS.RDF
         /// <summary>
         /// Creates a new instance of a Graph.
         /// </summary>
-        public Graph() 
+        public Graph()
             : base() { }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


kal:
Thanks @langsamu!

PR #329, opened 2020-07-27
fix: n-triples formatter double encoded escaped characters

tpluscode:
This should address the double-escaping problem and should also fix a problem that Turtle formatter escaped Unicode characters

Unfortunately there are insufficient tests or maybe I missed them because I run Rider on Mac. Let's see what CI says.

### DIFF
--- a/Libraries/dotNetRDF/Core/URINode.cs
+++ b/Libraries/dotNetRDF/Core/URINode.cs
@@ -2,21 +2,21 @@
 // <copyright>
 // dotNetRDF is free and open source software licensed under the MIT License
 // -------------------------------------------------------------------------
-// 
+//
 // Copyright (c) 2009-2020 dotNetRDF Project (http://dotnetrdf.org/)
-// 
+//
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
 // in the Software without restriction, including without limitation the rights
 // to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
 // copies of the Software, and to permit persons to whom the Software is furnished
 // to do so, subject to the following conditions:
-// 
+//
 // The above copyright notice and this permission notice shall be included in all
 // copies or substantial portions of the Software.
-// 
-// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR 
-// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, 
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
 // FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
 // AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY,
 // WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
@@ -25,6 +25,7 @@
 */
 
 using System;
+using System.Diagnostics;
 using System.Linq;
 using System.Runtime.Serialization;
 using System.Xml;
@@ -41,7 +42,8 @@ namespace VDS.RDF
 #if !NETCORE
     [Serializable,XmlRoot(ElementName="uri")]
 #endif
-    public abstract class BaseUriNode 
+    [DebuggerDisplay("{" + nameof(_uri) + "}")]
+    public abstract class BaseUriNode
         : BaseNode, IUriNode, IEquatable<BaseUriNode>, IComparable<BaseUriNode>, IValuedNode
     {
         private Uri _uri;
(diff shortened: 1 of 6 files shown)


Libraries/dotNetRDF/Writing/Formatting/BaseFormatter.cs
```diff
@@ -130,7 +130,7 @@ public virtual String FormatUri(String u)
         /// <returns></returns>
         public virtual String FormatUri(Uri u)
         {
-            return FormatUri(u.AbsoluteUri);
+            return FormatUri(u.ToString());
```
tpluscode, line 133: This is where Unicode character were unnecessarily escaped because `AbsoluteUri` is always percent-encoded

Libraries/dotNetRDF/Core/URINode.cs
```diff
@@ -41,7 +42,8 @@ namespace VDS.RDF
 #if !NETCORE
     [Serializable,XmlRoot(ElementName="uri")]
 #endif
-    public abstract class BaseUriNode 
+    [DebuggerDisplay("{" + nameof(_uri) + "}")]
```
tpluscode, line 45: Slightly unrelated. Can revert if you prefer

Libraries/dotNetRDF/Writing/Formatting/BaseFormatter.cs
```diff
@@ -118,9 +118,7 @@ public virtual String Format(Triple t)
         /// <returns></returns>
         public virtual String FormatUri(String u)
         {
-            // String uri = Uri.EscapeUriString(u);
-            u = u.Replace(">", "\\>");
-            return u;
+            return Rfc3987Formatter.EscapeUriString(u);
```
tpluscode, line 121: Interesting how it was `Uri.EscapeUriString` originally.

What was the reason actually to implement the `Rfc3987Formatter`. `System.Uri` is not conformant?

Libraries/dotNetRDF/Writing/Formatting/Rfc3987Formatter.cs
```diff
@@ -33,57 +33,97 @@ internal static class Rfc3987Formatter
         public static string EscapeUriString(string uriString)
         {
             var builder = new StringBuilder();
-            foreach (var c in uriString)
+            using (var enumerator = uriString.GetEnumerator())
```
tpluscode, line 36: By changing to enumerator I can peek the two chars following a `%` and append verbatim. The only reason to do otherwise is when the `%` is not followed by a valid escape sequence, dangling at the end of the string

Testing/unittest/Writing/WriterTests.cs
```diff
@@ -175,8 +175,25 @@ public void WritingXmlAmpersandEscaping()
             }
         }
 
-        [Fact]
-        public void WritingUriEscaping()
+        [Theory]
+        [MemberData(nameof(RoundTripTestData), Formats.All ^ Formats.RdfXml)]
```
tpluscode, line 179: RDF/XML does not do Unicode so excluded from this theory

Libraries/dotNetRDF/Writing/Formatting/BaseFormatter.cs
kal: IIRC there were some cross-platform issues. .NET Core and later .NET Framework properly implement IRI support. Earlier versions didn't. Looking at the docs (https://docs.microsoft.com/en-us/dotnet/api/system.uri?view=netcore-3.1), this appears to have been fixed - though .NET 4.0 users might have to fiddle with app.config to enable IRI support.

Might be worth trying to remove this formatter class - it's not 100% correct in its approach in any case and it would be less code to manage.

Libraries/dotNetRDF/Core/URINode.cs
kal: Fine with me!

Libraries/dotNetRDF/Writing/Formatting/BaseFormatter.cs
tpluscode: The tests are passing on net452. Are we even going to support older versions? I think it hardly makes sense even for <net462, correct?

> it's not 100% correct in its approach in any case and it would be less code to manage.

Maybe it is correct now? In any case, less code is always better

tpluscode: Ah, no luck with `Uri.EscapeUriString'. It suffers from the same problem if presented with a partially escaped string

```
// will also double-escape
Uri.EscapeUriString("http://example.com/Person/2/Bob/September%2C 2010")

// will leave coma (reserved char)
Uri.EscapeUriString("http://example.com/Person/2/Bob/September, 2010") 
```

kal: Ah well, it was worth a try, thanks :-)

Yeah support for older versions of .NET is definitely something to discuss - but probably a bigger issue than this PR!

tpluscode: Aha, the main binary is indeed compiled to `net40`. Then maybe unit tests should also run against that platform?

kal: Yeah, ideally it would. I just worry that adding a third target framework to the unit tests will cause us to hit Appveyor build time limits.

tpluscode: I have been quite happy with GitHub Actions in many of my repos. They are very fast (definitely faster then Travis, have not enough to compare with Appveyor)

And a big plus is ability to spread a build across multiple jobs which run in parallel. For example a job per target

kal: Created #330 to remind myself to take a look at some point!

tpluscode:
Awesome. Thanks for the release @kal. Now I'm only 2 test cases away from full R2RML compatibility 👌

PR #333, opened 2020-08-11
Fuzzy match media type in Link headers

tmarkovski:
Attempt to match mediaType in Link header with fuzzy match. I discovered this bug when using `https://schema.org` as context. The returned Link header has media type set to `"application/ld+json"`, while the code is looking for exact match `application/ld+json`.

### DIFF
--- a/Libraries/dotNetRDF/JsonLd/DefaultDocumentLoader.cs
+++ b/Libraries/dotNetRDF/JsonLd/DefaultDocumentLoader.cs
@@ -60,7 +60,7 @@ namespace VDS.RDF.JsonLd
             {
                 var contextLinks = ParseLinkHeaders(client.ResponseHeaders.GetValues("Link"));
                 var alternateLink = contextLinks.FirstOrDefault(link =>
-                    link.RelationTypes.Contains("alternate") && link.MediaTypes.Contains("application/ld+json", new MediaTypeComparer()));
+                    link.RelationTypes.Contains("alternate") && link.MediaTypes.Contains("application/ld+json"));
                 if (alternateLink != null)
                 {
                     return LoadJson(new Uri(remoteRef, alternateLink.LinkValue), loaderOptions);
@@ -116,6 +116,7 @@ namespace VDS.RDF.JsonLd
                         }
                         else if (key.Equals("type"))
                         {
+                            value = value.Trim('"');
                             mediaTypes.Add(value);
                         }
                     }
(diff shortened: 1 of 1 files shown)


tpluscode:
I don't think that's the best way to fix that. Isn't the problem somewhere in the code which parses the headers?

tmarkovski:
@tpluscode Thanks, you're right. I added the change in the parsing code.

tpluscode:
Great.
Curious though, is that a case where the HTTP response is incorrectly formatted?

tmarkovski:
I'm not sure, it seems quotes must be stripped on all parameters. Here's the schema org header response.

`link: </docs/jsonldcontext.jsonld>; rel="alternate"; type="application/ld+json"`

```
> curl -I https://schema.org
HTTP/2 200 
access-control-allow-credentials: true
access-control-allow-headers: Accept
access-control-allow-methods: GET
access-control-allow-origin: *
access-control-expose-headers: Link
link: </docs/jsonldcontext.jsonld>; rel="alternate"; type="application/ld+json"
date: Tue, 11 Aug 2020 17:10:25 GMT
expires: Tue, 11 Aug 2020 17:20:25 GMT
cache-control: public, max-age=600
etag: "Aif3Ow"
x-cloud-trace-context: eea5391c8685ce95ed5f82aef82c65b8
content-type: text/html
server: Google Frontend
content-length: 5100
alt-svc: h3-29=":443"; ma=2592000,h3-27=":443"; ma=2592000,h3-T050=":443"; ma=2592000,h3-Q050=":443"; ma=2592000,h3-Q046=":443"; ma=2592000,h3-Q043=":443"; ma=2592000,quic=":443"; ma=2592000; v="46,43"
```

kal:
> 
> 
> Great.
> Curious though, is that a case where the HTTP response is incorrectly formatted?

Looks like the spec allows the parameters to be quoted strings - both type and rel can be quoted (although neither *has* to be quoted). It also looks like rel can be a quoted, space-separated list of relation types so that might be another corner case that has to be handled...just in case.

tmarkovski:
> > Great.
> > Curious though, is that a case where the HTTP response is incorrectly formatted?
> 
> Looks like the spec allows the parameters to be quoted strings - both type and rel can be quoted (although neither _has_ to be quoted). It also looks like rel can be a quoted, space-separated list of relation types so that might be another corner case that has to be handled...just in case.

It seems that case is already handled in the code

```
                        if (key.Equals("rel"))
                        {
                            value = value.Trim('"');
                            relTypes.AddRange(value.Split(' '));
                        }
```

kal:
So I relied on the "Open this in GitHub Desktop" link to do the right thing, which of course it failed to do. So there is now a separate PR to move this forwards.

Thanks for the contribution @tmarkovski and the review @tpluscode ! #334 adds some unit tests which should hopefully make codecov happy again!

kal:
Merged #334 which includes this fix. Thanks @tmarkovski !

PR #335, opened 2020-08-17
Set HasBase in JsonLdContext only if Uri is not null

tmarkovski:
I discovered this bug trying to run the compaction algorithm against the below input with context `{}`. The exception I got was a null reference at [this line](https://github.com/dotnetrdf/dotnetrdf/blob/fb3f801130a23735bca60408c37a7da32af169c7/Libraries/dotNetRDF/JsonLd/Processors/CompactProcessor.cs#L1013) where `activeContext.Base` is null. It seems that `if` condition above it for `activeContext.HasBase` should have returned false.
I'm not sure how a unit test can be written for this, and if one is warranted.

JSON to reproduce this. [Permalink to JsonLd Playground](https://json-ld.org/playground/#startTab=tab-compacted&json-ld=%7B%22%40context%22%3A%5B%22https%3A%2F%2Fwww.w3.org%2F2018%2Fcredentials%2Fv1%22%2C%22https%3A%2F%2Fw3id.org%2Fcitizenship%2Fv1%22%5D%2C%22id%22%3A%22https%3A%2F%2Fissuer.oidp.uscis.gov%2Fcredentials%2F83627465%22%2C%22type%22%3A%5B%22VerifiableCredential%22%2C%22PermanentResidentCard%22%5D%2C%22issuer%22%3A%22did%3Aexample%3A489398593%22%2C%22identifier%22%3A%2283627465%22%2C%22name%22%3A%22Permanent%20Resident%20Card%22%2C%22description%22%3A%22Government%20of%20Example%20Permanent%20Resident%20Card.%22%2C%22issuanceDate%22%3A%222019-12-03T12%3A19%3A52Z%22%2C%22expirationDate%22%3A%222029-12-03T12%3A19%3A52Z%22%2C%22credentialSubject%22%3A%7B%22id%22%3A%22did%3Aexample%3Ab34ca6cd37bbf23%22%2C%22type%22%3A%5B%22PermanentResident%22%2C%22Person%22%5D%2C%22givenName%22%3A%22JOHN%22%2C%22familyName%22%3A%22SMITH%22%2C%22gender%22%3A%22Male%22%2C%22image%22%3A%22data%3Aimage%2Fpng%3Bbase64%2CiVBORw0KGgokJggg%3D%3D%22%2C%22residentSince%22%3A%222015-01-01%22%2C%22lprCategory%22%3A%22C09%22%2C%22lprNumber%22%3A%22999-999-999%22%2C%22commuterClassification%22%3A%22C1%22%2C%22birthCountry%22%3A%22Bahamas%22%2C%22birthDate%22%3A%221958-07-17%22%7D%7D&context=%7B%7D)
```json
{
  "@context": [
    "https://www.w3.org/2018/credentials/v1",
    "https://w3id.org/citizenship/v1"
  ],
  "id": "https://issuer.oidp.uscis.gov/credentials/83627465",
  "type": [ "VerifiableCredential", "PermanentResidentCard" ],
  "issuer": "did:example:489398593",
  "identifier": "83627465",
  "name": "Permanent Resident Card",
  "description": "Government of Example Permanent Resident Card.",
  "issuanceDate": "2019-12-03T12:19:52Z",
  "expirationDate": "2029-12-03T12:19:52Z",
  "credentialSubject": {
    "id": "did:example:b34ca6cd37bbf23",
    "type": [ "PermanentResident", "Person" ],
    "givenName": "JOHN",
    "familyName": "SMITH",
    "gender": "Male",
    "image": "data:image/png;base64,iVBORw0KGgokJggg==",
    "residentSince": "2015-01-01",
    "lprCategory": "C09",
    "lprNumber": "999-999-999",
    "commuterClassification": "C1",
    "birthCountry": "Bahamas",
    "birthDate": "1958-07-17"
  }
}

```

### DIFF
--- a/Libraries/dotNetRDF/JsonLd/JsonLdContext.cs
+++ b/Libraries/dotNetRDF/JsonLd/JsonLdContext.cs
@@ -71,7 +71,7 @@ namespace VDS.RDF.JsonLd
         /// <remarks>The value may be a relative or an absolute IRI or null.</remarks>
         public Uri Base { 
             get => _base;
-            set { _base = value; HasBase = true; }
+            set { _base = value; HasBase = value != null; }
         }
 
         /// <summary>
--- a/Libraries/dotNetRDF/JsonLd/JsonLdContext.cs
+++ b/Libraries/dotNetRDF/JsonLd/JsonLdContext.cs
@@ -71,13 +71,8 @@ namespace VDS.RDF.JsonLd
         /// <remarks>The value may be a relative or an absolute IRI or null.</remarks>
         public Uri Base { 
             get => _base;
-            set { _base = value; HasBase = value != null; }
+            set { _base = value; }
         }
-
-        /// <summary>
-        /// Returns true if the Base property of this context has been explicitly set.
-        /// </summary>
-        public bool HasBase { get; private set; }
 
         /// <summary>
         /// Get the base IRI that this context was originally created with.
@@ -140,7 +135,6 @@ namespace VDS.RDF.JsonLd
         public void RemoveBase()
         {
             Base = null;
-            HasBase = false;
         }
 
         /// <summary>
@@ -152,7 +146,6 @@ namespace VDS.RDF.JsonLd
             var clone = new JsonLdContext(OriginalBase)
             {
                 Base = Base,
-                HasBase = HasBase,
                 Language = Language,
                 BaseDirection = BaseDirection,
                 Version = Version,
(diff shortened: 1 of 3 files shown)


kal:
The definition of the HasBase property is that it should return true if the Base property was explicitly set to null, but in the code the only 2 places it gets used are also assuming that means that if context.HasBase is true then context.Base is not null. IIRC the reason I originally added that property with the definition it has is that it is possible for one to explicitly set the base to null, so that ends up overloading the meaning of null if we use it both for a value that is not set and for a value that has been explicitly set to null. That said, that explicit setting of a null context is not really an issue in the code-base and those two places where the test uses HasBase could just as easily  use context.Base != null and have the same semantics.

So perhaps for clarity it would be better to just drop `HasBase` entirely and rewrite the two places where it is used to just test for a non-null context.Base value. That would (a) simplify the code a bit and (b) avoid others making the same mistake with the context API :-)

tmarkovski:
@kal Makes sense. Pushed an update based on comments.

tmarkovski:
It looks like removing some code reduced the coverage % by 0.0.1 and codecov doesn't like that.

kal:
Yeah, I think I'm going to ignore codecov on this one! :)

Issue #340, opened 2020-10-09
Performance issue retracting triples from large graph


blackwork:
Per default, the TreeIndexedTripleCollection implementation is used to manage a tree index for a graph. There is a performance issue retracting triples from a large graph.

STR:
- create graph and add ~500.000 triples
- select e.g. ~20.000 triples from the graph
- retract the list of triples from the graph

To manage the index, `List<Triple>` is used. In Remove situations, the `Contains(triple)` method if .NET `List` has a poor performance. 
Switch to `HashSet<Triple>` will increase the performance.

kal:
Closing this issue as the fix was merged.

PR #341, opened 2020-10-09
Performance improvement retract triples using HashSet in TreeIndexedTripleCollection

blackwork:
Fixes issue #340

Replacing List<Triple> by HashSet<Triple> will increase the retract performance.


kal:
Hi @blackwork!

First of all, thank you for the PR! The code changes all look good to me. 

However, my one concern with this change is the potential memory overhead of the HashSet implementation. I'd probably need a bit of time to do some side-by-side comparisons on this, unless that is something you have already looked into?

kal:
Thanks @blackwork! Yeah, that doesn't look too bad for overhead and this will definitely make it easier to work with larger graphs in memory!

PR #356, opened 2020-12-10
Feature/issue 317

kal:
This PR makes two major updates to the API.

1. Adds a Name property to the IGraph interface that is now used as the graph name **instead of** the BaseURI property. Unlike BaseURI, the Name property is readonly and can only be set via the constructor.

1. Removes the need for every node to have a Graph context. The Graph property on INode is still there, but the goal will be to remove it from the interface (possibly with the exception of the IBlankNode interface to retain the ability to differentiate between two blank nodes with the same internal ID and different graph sources).

This is a major API change and although it doesn't break much in the public APIs, it does introduce a breaking change to the behaviour around named graphs.


kal:
@langsamu, @tpluscode - just tagging you on this PR in case you want to take a look before I merge. It's alot of change that basically all flowed out of allowing graphs to have a blank node as a name and the knock-on effects that this means because we were using BaseUri to serve as both the graph name and a default base URI for relative URI resolution. It also resulted in the logical course of action being to remove the Graph property from the INode interface - this was in use in a couple of places...most especially in the SHACL processor where it has basically meant adding an additional parameter to a lot of internal constructors but doesn't really change anything in terms of the SHACL API. 

The biggest changes from a user perspective are that you set the graph name in the constructor, not by changing the BaseUri property. So a typical pattern of:

    var g = new Graph();
    g.Load(...);
    g.BaseUri = new Uri("http://example.org/graph");

becomes:

    var g = new Graph(new UriNode(new Uri("http://example.org/graph")));
    g.Load(...);

The constructor that takes an IRefNode would allow you to give the graph a blank node name. There is also an overload for the constructor that takes a URI so the above can be shortened to:

    var g = new Graph(new Uri("http://example.org/graph"));
    g.Load(...);

There is still some tidying up to do - I think for example that it now makes sense to refactor IGraph and its implementations to not derive from INodeFactory and have those as two separate concepts. INodeFactory might also be a better place to put BaseUri as URI resolution maybe makes more sense in the context of creating UriNodes. So definitely a lot of additional work to do to streamline and clean up the APIs but I hope you guys see this as a good step to take!

Issue #382, opened 2021-03-22
optimization proposal for Service


giacomociti:
Hello, I'm experimenting with an optimization based on [this W3C document](https://www.w3.org/TR/2013/REC-sparql11-federated-query-20130321/#values). 
Instead of calling the remote query once for each solution, we can inject all the values at once (or in chunks of limited size).
If this is of any interest I can provide a PR.

kal:
Hi Giacomo - that does sound interesting and a useful way to optimise the evaluation of the Service algebra. Are you planning to implement this by updating the code in Libraries\dotNetRDF\Query\Algebra\Service.cs?

giacomociti:
Yes, I just opened a PR so you can have a look at the code

PR #383, opened 2021-03-23
Service optimization proposal

giacomociti:
implementation for proposal #382

### DIFF
--- a/Testing/dotNetRDF.MockServerTests/SparqlServiceTests.cs
+++ b/Testing/dotNetRDF.MockServerTests/SparqlServiceTests.cs
@@ -1,0 +1,195 @@
+/*
+dotNetRDF is free and open source software licensed under the MIT License
+
+-----------------------------------------------------------------------------
+
+Copyright (c) 2009-2017 dotNetRDF Project (dotnetrdf-developer@lists.sf.net)
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is furnished
+to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR 
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, 
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY,
+WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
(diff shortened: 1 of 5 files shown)


giacomociti:
testing against a local GraphDB instance (too often I could not reach DBPedia) I noticed that the intended heuristics for guessing Uris in CSV was not working.

giacomociti:
This PR definitively needs more testing (either automatic or manual).
Sending longer queries to remote endpoints affects many heuristics (GET vs POST etc.) so we need to ensure that this is reasonably robust in practice.

kal:
Hi @giacomociti ,

This is a great start though! I've got to spend a bit more time reviewing the changes in Service.cs in a code editor because the restructuring you did (which is good!) makes it a bit hard to see exactly what's going on, but I think I get the general direction.

With regards to testing, I started working on moving a lot of the tests we had that depended on external resources into a separate test library that uses WireMock.Net to provide mock endpoints. I think you could do that here. The FederatedEndpointFixture sets up a couple of mock servers with endpoints that do various things - I'm not sure that fixture is exactly what we need for testing the different pathways through the code, but it maybe provides a good starting point for creating a useful test fixture for testing the Service code?

BTW was the fix for the URI guessing because of the content returned from the dbpedia SPARQL endpoint? I'm just curious as to why the SPARQL CSV parser is involved as I would have thought that the default return from a SPARQL endpoint would always be SPARQL XML...

giacomociti:
Hi @kal ,
I got CSV results when using a local GraphDB instance as the endpoint.
When using DBPedia instead I got also some strange error: (406) Unacceptable.

That's why I think that, while WireMock is certainly valuable for tests and I will try to use it, we also need to check the behavior with common triplestores.

kal:
I agree it is worth including tests target the other triplestores if possible. That's a bit hard in the current setup because there's no common test environment. On the 3.0 branch I've done a lot of work on breaking out the triplestore-specific tests and setting up a dockerfiles that can build containers for running the tests against the FOSS triplestores that we support. Its not something I want to backport to the 2.x branch right now though.

So I'd say lets put in place WireMock-based tests as the unit tests for the SERVICE feature (including things like testing that we are sending meaningful Content-Type and Accept headers for example) and then track other issues as things to investigate separately (perhaps as separate tickets) - especially where the problem might actually be to do with a store-specific implementation of the SPARQL protocols. For those store-specific tests we can implement them on the 2.x branch with the current pattern and as I make more progress on the 3.0 work I'll port any new triple-store specific tests we create over there.

Does that sound OK?

Testing/dotNetRDF.MockServerTests/SparqlServiceTests.cs
```diff
@@ -0,0 +1,223 @@
...
+        public void ItResolvesNoEndpointFromVariable()
+        {
+            var dataset = CreateDataset(numberOfTriples: 0);
+            var query = $"SELECT * WHERE {{ VALUES ?service {{ 'noturi' }} SERVICE ?service {{?s ?p ?o}} }}";
+
+            var results = ProcessQuery(dataset, query);
+
+            results.Should().NotBeNull().And.HaveCount(0);
+        }
+
```
kal, line 185: That's interesting! This matches the logic that we had for SERVICE previously so I'm not complaining, but I wonder if this behaviour is slightly at odds with the specification as https://www.w3.org/TR/2013/REC-sparql11-federated-query-20130321/#serviceFailure says that a query should fail if the endpoint IRI cannot be resolved (though it doesn't say what to do if the SERVICE endpoint value is not an IRI).

In any case I'm happy to keep the behaviour we had until someone complains :-)

kal (approved):
This looks great, thank you @giacomociti ! It's a much cleaner implementation now too. 👍

Issue #384, opened 2021-03-24
Feature Request: break down assemblies


jiatao99:
Is it possible to follow .NET core design pattern to break down the big dotnetrdf assembly into smaller assemblies? Something like:

dotnetrdf.abstraction.dll:  interfaces
dotnetrdf.client.dll: client invoking library including QueryBuilder

Most user just need the two libraries above to connect external RDF database

dotnetrdf.parser.dll: SPARQL parser
dotnetrdf.server.dll: in memory RDF implementation

I know it is going to be a huge refactor, but I think it worth the effort to make it more efficient. 

What do you think?

kal:
That was something I had thought about for the 3.0 release and it has been quite a long-standing issue (see https://github.com/dotnetrdf/dotnetrdf/issues/32 - the 1.9 that is referred to in there is work that Rob did a long time ago on a refactor but that work didn't end up being our 2.0 release). I think it is worth revisiting and thinking about the split that we would take. I think there are roughly these elements:

- Core Abstractions
- Parsing & Serialization
- In-memory store and in-memory SPARQL engine
- Standard protocol clients (SPARQL protocol mainly)
- Third party triple store integrations (a separate assembly for each integration?)
- Dynamic Graphs
- SPARQL Query Builder
- SPIN library
- Full-text indexing

Some of these might need to go together or make sense to go together so it might not end up being that many separate assemblies (for example your suggestion that SPARQL query builder should be included in a client library that would  also cover the remote protocols)

One thing to guide how we split things up would be to decided exactly why we are doing that. One reason is to make the size of install smaller. Another is to reduce the number of dependency libraries that are called in. So if isolating dependencies is an issue we might actually divide things up differently (for example the support for RDFa pulls in Anglesharp / HTMLAgilityPack so we could isolate the RDFa parser/serialiser into its own package so that dependency is also isolated).

What would your motivation be for doing this break down?

kal:
Just for reference, this is the document that Rob created when planning out the 1.9 design with his original plan of the breakdown of modules (it doesn't cover everything that is in the library now though): https://github.com/dotnetrdf/dotnetrdf/wiki/DeveloperGuide-1.9-Libraries

jiatao99:
Thanks. That's a very good document. A couple of suggestion:

[...]

PR #429, opened 2021-09-26
Release 3.0 - initial work

kal:
I'm opening this PR for comments. Although the work on the 3.0 release is not done yet, there are already a lot of breaking changes piled up on the release/3.0 branch and I think it would be good to get some feedback before moving on to the main feature development (which will be RDF-* support).

Probably the most significant thing is the re-organisation of the libraries. In response to #384 I've broken the dotNetRDF assembly down into a core assembly for the model, parsers and serialisers; a client assembly for the store connectors; and an in-memory assembly for the in-memory store and Leviathan query engine. I've also split out other self-contained units of functionality into their own assemblies (e.g SHACL support, SKOS support, Inferencing etc.). The projects are set up so that each of these assemblies is also a separate NuGet package so it should be easier for devs to pull in just the dependencies they need.

The other significant piece of work was around the core model for Triples and nodes and making both triples and nodes  of the graph(s) that contain them. This makes it much easier to assert triples in multiple graphs and removes a lot of the triple/node copying code. This will be helpful when getting to the RDF-* work as we will want triples to be allows as nodes.

There is also a change to naming graphs - a graph can now be named by an IUriNode or IBlankNode. This name is now set in the constructor of the graph and is immutable. This also clears up the confusion we had between the BaseUri for the graph (used for resolving relative URI references) and the graph name.

I've also done quite a bit of work on trying to eliminate the large number of static options and move them into constructor-injected options instead. This had quite a big ripple effect through the code-base, but it does make it much easier to localise changes to options and if a user really wants to set up some singleton instances then the constructor-injection should make it easy to do that either manually or through an IoC library.

As you'll see from the helpful message above, this change is (a) massive and (b) not automatically mergeable with master. I don't expect that the GitHub diff tools will be helpful for reviewing this stuff so I'd recommend any interested reviewer instead just check out the release/3.0.0 branch and build it locally.


kal:
Closed this PR as I'm doing the update by branch renaming as described on #453

PR #533, opened 2022-12-26
Fix Docker compose

stefannikolei:
With this changes `docker compose up` is working again. 

This makes testing locally easier.


kal:
Thanks @stefannikolei!

stefannikolei:
@kal should all tests pass when these containers are running. 

Can test it on windows and Mac. 

Under Mac many are failing.

kal:
@stefannikolei Ideally the tests should pass on the Mac, but I don't have access to one to be able to build and test on that environment. I usually build and test locally under Windows, but the connector tests that we run in the CI build are running under Ubuntu. so if there are problems on the Mac it might be more related to networking that to line endings (??). 

If you can get some information about what tests are failing and any relavant stack traces or error messages that would be really helpful. It would be best to open that as a separate issue from this PR though if that's OK.

PR #537, opened 2023-01-02
Drop .net472 as target

stefannikolei:
* removed the Projects which only targetted net472 (Virtuoso and ASP.NET)
* removed all conditions in csproj
* Removed ConfigurationManagerSettingsProvider (To use ISettingsprovider a new Implementation must be written (probably best with IConfiguration))
* hopefully configured the build in the correct way :-)


kal:
Hi @stefannikolei! Thanks for the PR. However I just merged my changes to implement this - basically the same set of changes but keeping net472 as a test framework that is supported.

PR #540, opened 2023-02-05
Drop libraries that are .NET Framework specific

kal:
* Remove dotNetRdf.AspNet
* Remove dotNetRdf.Connectors.Virtuoso
* Remove the corresponding test libraries
* Refactor Directory.build.props to have separate props for libraries and tests
* Update all libraries to target netstandard2.0
* Update all tests to target net6.0, net7.0 and net472
* Update GH Actions test matrix to cover net6.0, net7.0 and net472


kal:
This PR is intended as an alternative to #537. The main difference being that the libraries all target NET Standard 2.0 to support a wider range of target platforms that NET6 and NET7 as on #537.  With this approach NET Framework is still supported as a runtime framework and so test libraries should still include it as a target.

Issue #560, opened 2023-04-13
Formatting/handling comments


IS4Code:
I have a situation where I need to add logging information to the serialized RDF output, but there is no standard way to represent a handler that can process comments.

I could use something like this:
```cs
namespace VDS.RDF.Writing.Formatting
{
    public interface ICommentFormatter
    {
        string FormatComment(string text);
    }
}
```
Formatters could implement this interface, for example `TurtleFormatter` returning `"#text"`, `RdfXmlFormatter` returning `<!--text-->` etc. (with proper sanitization)

In addition, there could be a general way to inject comments into RDF data, such as:
```cs
namespace VDS.RDF
{
    public interface ICommentRdfHandler : IRdfHandler
    {
        bool HandleComment(string text);
    }
}
```
An implementation like `WriteThroughHandler` could then query the formatter from this method for the `ICommentFormatter` interface and write the comment line to the output.

What do you think about such an addition? I can implement this myself through a pull request, if needed.

kal:
I think this sounds like a good extension to add. The design with the separate `ICommentFormatter` interface makes sense. 
Implementations would have to handle both string sanitization and also handling multi-line comments, but that logic properly belongs in the formatter implementations.

If we had `ICommentRdfHandler` as an extension of `IRdfHandler` then we could also potentially later extend some of the parsers to report comment strings, which might also be a useful feature. I wonder about just adding HandleComment to the existing `IRdfHandler` interface as a simplification, but the trouble with that is that it would introduce a breaking API change as all `IRdfHandler` implementations would have to be updated. So on balance `ICommentRdfHandler` is probably the way to go.

A PR to add this would be welcome if you have the time!

kal:
Thanks for the PR! I've added this to the 3.1 milestone, but don't have a release date in mind, so please let me know if it is important for you to have "official" NuGet packages available.

IS4Code:
Thank you too! I am in no hurry at the moment, so it can wait a few months for sure. 😃

PR #561, opened 2023-04-13
ICommentFormatter and ICommentRdfHandler

IS4Code:
Per #560

### DIFF
--- a/Libraries/dotNetRdf.Core/Writing/Formatting/NTriplesFormatter.cs
+++ b/Libraries/dotNetRdf.Core/Writing/Formatting/NTriplesFormatter.cs
@@ -27,6 +27,7 @@
 using System;
 using System.Collections.Generic;
 using System.Text;
+using System.Text.RegularExpressions;
 using VDS.RDF.Parsing;
 
 namespace VDS.RDF.Writing.Formatting
@@ -35,7 +36,7 @@ namespace VDS.RDF.Writing.Formatting
     /// Formatter for formatting as NTriples.
     /// </summary>
     public class NTriplesFormatter
-        : BaseFormatter
+        : BaseFormatter, ICommentFormatter
     {
         private readonly BlankNodeOutputMapper _bnodeMapper;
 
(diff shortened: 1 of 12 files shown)


Libraries/dotNetRdf.Core/Writing/Formatting/NTriplesFormatter.cs
```diff
@@ -278,6 +279,17 @@ public override string FormatUri(string u)
...
+
+        /// <summary>
+        /// Matches the beginning of all lines.
+        /// </summary>
+        static readonly Regex commentLineRegex = new Regex(@"^", RegexOptions.Compiled | RegexOptions.Multiline);
+
+        /// <inheritdoc />
+        public virtual string FormatComment(string text)
+        {
+            return commentLineRegex.Replace(text, "#");
```
kal, line 291: Might be nicer to format the comment with a space after the `#` ?

Libraries/dotNetRdf.Core/Writing/WriterUtilities.cs
```diff
@@ -769,6 +771,42 @@ public static string EncodeForXml(string value)
...
         }
 
+        /// <summary>
+        /// Removes characters invalid in XML documents.
+        /// </summary>
+        /// <param name="content">The string to filter.</param>
+        /// <returns>
+        /// New string that does not contain any invalid characters.
+        /// </returns>
+        public static string RemoveInvalidXmlChars(string content)
```
kal, line 781: I wonder if it would be easier to just filter on the XML definition of a Char which is:
```
#x9 | #xA | #xD | [#x20-#xD7FF] | [#xE000-#xFFFD] | [#x10000-#x10FFFF]
```
and then drop characters that are not in one of those ranges. It would avoid having to handle multiple exceptions in those corner cases where a comment has lots of disallowed chars.

kal (changes requested):
Looks good! Just a couple of minor comments/suggestions for change.

Libraries/dotNetRdf.Core/Writing/Formatting/NTriplesFormatter.cs
kal: That's a good point, and as long as we are consistent in that approach in all of the formatters then I think it would be enough just to document it.

Libraries/dotNetRdf.Core/Writing/WriterUtilities.cs
kal: Yeah, I think that would be a viable approach and would make it clear what sanitisation the formatter is doing.

kal (approved):

PR #570, opened 2023-05-23
Add JsonLdformatter and fix some minor things

IS4Code:
This brings several updates to formatting:
* `JsonLdWriter` implements `IFormatterBasedWriter` and uses `JsonLdFormatter` when creating `JObject` from `INode`.
* The new `JsonLdFormatter` supports writing "graphs" as JSON arrays of node objects, one from each triple, similarly to how the RDF/XML formatter does it.
* `WriteThroughHandler` is updated to consider both `_formatter` and `_tripleFormatter` when formatting other things than triples.
* The actual JSON-LD formatting has been fixed a bit ‒ when `UseNativeTypes` is set, but the lexical value is not valid for the datatype, the triple is correctly emitted anyway (the datatype was ignored previously), just with the original value as if no `UseNativeTypes` was set. The same for `rdf:JSON` which previously raised an exception when the value was invalid. Additionally, `XmlConvert.ToInt64` is used instead of `long.Parse`, and the same for `double` (for XSD-compatible culture-neutral parsing).

### DIFF
--- a/Libraries/dotNetRdf.Core/Writing/Formatting/JsonLdFormatter.cs
+++ b/Libraries/dotNetRdf.Core/Writing/Formatting/JsonLdFormatter.cs
@@ -142,7 +142,9 @@ namespace VDS.RDF.Writing.Formatting
             var node = FormatAsJObject(t);
             if (graph != null)
             {
-                node["@graph"] = FormatAsJObject(graph);
+                var graphNode = FormatAsJObject(graph);
+                graphNode["@graph"] = node;
+                node = graphNode;
             }
             return node.ToString(JsonFormatting) + ",";
         }


IS4Code:
The failing tests are from js08-in.nq and js09-in.nq, where a lexically invalid `rdf:JSON` is used. I could try to update that too if needed.

Perhaps some rationale is in order ‒ I think that JSON-LD, as an RDF *serialization* format, should not restrict what kind of RDF could be represented in it, especially when the same kind of data would be fine under JSON-LD 1.0. The fact that a specific literal value is not representable when using `@json` does not mean it cannot be represented at all, since it is nonetheless a valid literal under RDF. (Similarly in RDF/XML a `rdf:XMLLiteral` could easily be represented without using `parseType="Literal"`)

PR #608, opened 2023-11-12
Introducing Linked Data Fragments

### LINKED ISSUE #200: Linked Data Fragments
Is there any plan to had [Linked Data Fragments](http://linkeddatafragments.org/) client?

langsamu:
### Summary
This change resolves #200 and introduces a Triple Pattern Fragments (TPF) client to dotNetRDF.

The client is an implementation of the `IGraph` interface which dispatches graph operations to a TPF endpoint.

### Value
The advantage of this approach is that developers can interact with Linked Data Fragments in the exact same manner as they do with any other graph (constructed programmatically or loaded from disk or network).

A corollary is the ability to run SPARQL queries against a TPF endpoint. This is a feature that other LDF clients (like [Comunica](https://github.com/comunica/comunica)) achieve by mapping the SPARQL algebra to LDF operations. In our case this is achieved by the Leviathan query engine that maps the same to methods exposed by `IGraph`, which the implementation introduced here translates to TPF network calls.

### Limitations
I've chosen to implement TPF as opposed to Quad Pattern Fragments (QPF) because the chosen implementation approach is not viable for QPF in dotNetRDF as we don't really have quads. Of course QPF could be implemented in DNR in multiple ways, but I think there is still value in exploring this implementation paradigm.

This is a naive implementation. There are no robustness/resilience measures like the capability to rety and back off. It is very likely that production usage will suffer from network outage and throttling. For now, mitigating these scenarios is left to applications hosting this client.
This is an intentional choice: I'd rather start with a naive solution and allow usage patterns to drive further development. Better than me trying to introduce plumbing here based on assumption only.

The client introduced here is also naive in its lack of support for caching. Again, this is by design.

#### Mitigation
Same applies to the current lack of support for authentication, which is likely to be required in all but the most simple use-cases.

There is a limited extensibility mechanism introduced here that might help aleviate some of the above shortcomings: The client accepts an optional `Loader`. Callers could implement resilience, caching or authentication here.

### Architecture

This feature focuses on the consumption of TPF, so it is under a `Client` namespace.
Future work could implement server features.

- `VDS.RDF.LDF.Client.TpfLiveGraph` is the single public API of this client.
It's an `IGraph` whose triples are a `TpfTripleCollection`. This is its major contribution.
In addition it overrides several methods to advertise lack of support for mutation, blank nodes and RDF*.

- `VDS.RDF.LDF.Client.TpfTripleCollection` is the main mapping between our graph operations and LDF.
It's a `BaseTripleCollection`implemented by linking `WithXXX` and other methods to our `TpfEnumerable`.

- `VDS.RDF.LDF.Client.TpfEnumerator` (and `TpfEnumerable`) is the workhorse of the library.
It is responsible for returning triples from the TPF endpoint to our client.
It is the implementation of LDF paging.

- `VDS.RDF.LDF.Client.TpfLoader` is our representation of the TPF response.

- `VDS.RDF.LDF.Client.TpfMetadataGraph` is our representation of the TPF response metadata.

- `VDS.RDF.LDF.Client.TpfParameters` is our representation of a TPF request.


Libraries/dotNetRdf.LinkedDataFragments/dotNetRdf.LinkedDataFragments.csproj
```diff
@@ -0,0 +1,30 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <AssemblyName>dotNetRdf.LDF</AssemblyName>
+    <AssemblyTitle>dotNetRdf Linked Data Fragments Support</AssemblyTitle>
+    <RootNamespace>VDS.RDF.LDF</RootNamespace>
```
kal, line 6: I think I prefer this as the namespace. In the source files you have `VDS.RDF.LinkedPatternFragments` but I think either `VDS.RDF.LDF` or `VDS.RDF.LinkedDataFragments` would be better and in either case should be consistent with this project file declaration.

Libraries/dotNetRdf.LinkedDataFragments/Enumerable.cs
```diff
@@ -0,0 +1,46 @@
...
+// </copyright>
+*/
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace VDS.RDF.LinkedPatternFragments
+{
+    internal class Enumerable : IEnumerable<Triple>
```
kal, line 33: Maybe a more specific name like `LdfEnumeration` or `TripleEnumeration` would be better?

Libraries/dotNetRdf.LinkedDataFragments/Enumerator.cs
```diff
@@ -0,0 +1,83 @@
...
+// </copyright>
+*/
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace VDS.RDF.LinkedPatternFragments
+{
+    internal class Enumerator : IEnumerator<Triple>
```
kal, line 33: I'd also update the name of this class if renaming `Enumerable` e.g. to `LdfEnumerator` or `TripleEnumerator`

Libraries/dotNetRdf.LinkedDataFragments/Graph.cs
```diff
@@ -0,0 +1,93 @@
...
+// </copyright>
+*/
+
+using System;
+using System.Collections.Generic;
+using VDS.RDF.LinkedPatternFragments.Hydra;
+
+namespace VDS.RDF.LinkedPatternFragments
+{
+    public class Graph : RDF.Graph
```
kal, line 33: Can we call this `LdfGraph` (or something like that) so that it doesn't clash with the standard in-memory `Graph` implementation in the core library?

Libraries/dotNetRdf.LinkedDataFragments/Graph.cs
```diff
@@ -0,0 +1,93 @@
...
+                {
+                    mapping = new Dictionary<INode, INode>();
+                    return true;
+                }
+            }
+
+            return base.Equals(g, out mapping);
+        }
+
+        public override IBlankNode GetBlankNode(string nodeId)
```
kal, line 73: This implementation could just return null rather than throwing an exception.

Libraries/dotNetRdf.LinkedDataFragments/Parameters.cs
```diff
@@ -0,0 +1,102 @@
...
+        private readonly INode @object;
+
+        internal Parameters(IriTemplate template, INode subject = null, INode predicate = null, INode @object = null)
+        {
+            if (template is null)
+            {
+                throw new ArgumentNullException(nameof(template));
+            }
+
+            if (subject is object && subject.NodeType != NodeType.Uri)
```
kal, line 49: I think ` != null` is maybe clearer to read than `is object` ?

Libraries/dotNetRdf.LinkedDataFragments/TripleCollection.cs
```diff
@@ -0,0 +1,160 @@
...
+*/
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using VDS.RDF.LinkedPatternFragments.Hydra;
+
+namespace VDS.RDF.LinkedPatternFragments
+{
+    internal class TripleCollection : BaseTripleCollection
```
kal, line 34: Renaming this with a prefix to make it separate from the core library's TripleCollection class would be good.

Libraries/dotNetRdf.LinkedDataFragments/TripleStore.cs
```diff
@@ -0,0 +1,69 @@
...
+// </copyright>
+*/
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace VDS.RDF.LinkedPatternFragments
+{
+    internal class TripleStore : WrapperTripleStore
```
kal, line 33: Add a prefix to the name?

Testing/dotNetRdf.LinkedDataFragments.Tests/GraphTests.cs
```diff
@@ -0,0 +1,237 @@
...
+using System;
+using VDS.RDF.Parsing;
+using VDS.RDF.Query;
+using VDS.RDF.Writing.Formatting;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace VDS.RDF.LinkedPatternFragments
+{
+    public class GraphTests
```
kal, line 35: It would be good to change these tests to remove the dependency on an external service, perhaps using WireMock to mock the expected service responses. The tests should also have some assertions in them instead of / as well as the console output for use in the CI build. I'm happy to take on those tasks if you like!

kal (commented):
Hi @langsamu. As the PR is still in draft I thought I would just leave these comments for you. I'm happy to take on any/all of the suggested changes myself, but I don't want to step on your toes if there is still some further work you want to do on the PR before submitting it as ready for review.

langsamu:
Thanks for the initial review @kal.

This is far from done.
I was working on mock responses when I noticed your comment.
I'll definitely tackle naming as well.

Libraries/dotNetRdf.LinkedDataFragments/Parameters.cs
langsamu: 173a0bd

Testing/dotNetRdf.LinkedDataFragments.Tests/GraphTests.cs
langsamu: 8bb950c

Libraries/dotNetRdf.LinkedDataFragments/Graph.cs
langsamu: I keep toggling between `throw` and `null` for these.

I'd like the least misleading response, and an expressive exception seems best, because there's really no point asking one of these graphs to create a blank node (or quoted stuff).

But there's no harm, either, so I'm happy to modify if you think that style suits better.

What do you think about the methods that currently throw in [(Ldf)`TripleCollection`](https://github.com/dotnetrdf/dotnetrdf/pull/608/files#diff-a609b43b628dab518ea52fbe67ff7d112d60cc92f74de46863973404824b35e3)?

kal: Good point about TripleCollection! 

I think for the methods and readonly properties that are not intended to modify the graph (`GetBlankNode`) or TripleCollection (`QuotedCount`, `QuotedXxxNodes`,  `Quoted` , `ContainsQuoted`), we should return a default value (null, an empty enumeration, 0 or false depending on the property/method). For methods that are intended to modify the graph or collection we should throw a meaningful Exception.

Libraries/dotNetRdf.LinkedDataFragments/Parameters.cs
```diff
@@ -0,0 +1,102 @@
...
+namespace VDS.RDF.LinkedPatternFragments
+{
+    internal class Parameters
+    {
+        private readonly IriTemplate template;
+        private readonly INode subject;
+        private readonly INode predicate;
+        private readonly INode @object;
+
+        internal Parameters(IriTemplate template, INode subject = null, INode predicate = null, INode @object = null)
```
kal, line 42: Could we change the signature of this constructor to use IUriNode arguments. That would make the NodeType checks redundant and provide a more type-safe interface.

Libraries/dotNetRdf.LinkedDataFragments/Graph.cs
langsamu: aca6e70

Libraries/dotNetRdf.LinkedDataFragments/dotNetRdf.LinkedDataFragments.csproj
langsamu: 649d4b6

Libraries/dotNetRdf.LinkedDataFragments/Enumerable.cs
langsamu: Yes. This was fine for the prototype but needs distinction from the core classes.

a3ed684

Libraries/dotNetRdf.LinkedDataFragments/Enumerator.cs
langsamu: a3ed684

Libraries/dotNetRdf.LinkedDataFragments/Graph.cs
langsamu: a3ed684

Libraries/dotNetRdf.LinkedDataFragments/TripleCollection.cs
langsamu: a3ed684

Libraries/dotNetRdf.LinkedDataFragments/TripleStore.cs
langsamu: a3ed684

Libraries/dotNetRdf.LinkedDataFragments/Parameters.cs
langsamu: I don't think that would be better.

This constructor practically takes a (possibly truncated) triple, and our `VDS.RDF.Triple` allows all node types in all positions.
The `subject` and `predicate` parameters here come from places like `Triple.Subject` and `BaseTripleCollection.WithSubject` (etc.), all of which are `INode`s.

I think the generality + checks here are the best compromise.

kal: That's a really good point. I hadn't taken that into consideration in my initial review.

langsamu:
@kal I think this is ready for an interim review.

kal:
This is great, thanks @langsamu ! I've taken a read through the code and don't have any comments to add beyond the few things that you have already got noted in TODOs. FWIW I think that the use of SPARQL rather than SHACL is fine internally. Being able to configure the parser used would be good, but I don't feel that its absolutely critical to getting the feature merged (though if its not done then worth creating an issue as a reminder to get it done later). 

The only missing element for me now is documentation of public classes and methods and ideally also some overview documentation somewhere in the Users' Guide.

langsamu:
> This is great, thanks @langsamu ! I've taken a read through the code and don't have any comments to add beyond the few things that you have already got noted in TODOs. FWIW I think that the use of SPARQL rather than SHACL is fine internally. Being able to configure the parser used would be good, but I don't feel that its absolutely critical to getting the feature merged (though if its not done then worth creating an issue as a reminder to get it done later).
> 
> The only missing element for me now is documentation of public classes and methods and ideally also some overview documentation somewhere in the Users' Guide.

- f7104d9
- dde5584
- 293f510
- 12adb18

langsamu:
All yours @kal.

kal (approved):
Thanks for all of this work @langsamu! This is a really useful addition to the library and I'm really happy to approve! 🎉

Issue #615, opened 2024-01-01
RDF Dataset Canonicalization


veikkoeeva:
I did not see an issue regarding this, so I record to make this visible. I hope this an OK way of doing this.

In W3C verifiable credentials linked proofs use  [RDF Dataset Canonicalization](https://w3c.github.io/rdf-canon/spec/). From the specification, the relevant part seem to be [§ 5. Serialization](https://w3c.github.io/rdf-canon/spec/#serialization). To my fairly untrained mind it looks like there are some smallish changes to n-quads, e.g.

[Example 8](https://w3c.github.io/rdf-canon/spec/#ex-ser-unique-hashes): Canonical N-Quads representation of Unique Hashes
```
This example illustrates the result of serializing the [canonicalized dataset](https://w3c.github.io/rdf-canon/spec/#dfn-canonicalized-dataset) described in [Example 2](https://w3c.github.io/rdf-canon/spec/#ex-ca-unique-hashes).

<http://example.com/p> <http://example.com/q> _:c14n0 .
<http://example.com/p> <http://example.com/r> _:c14n1 .
_:c14n0 <http://example.com/s> <http://example.com/u> .
_:c14n1 <http://example.com/t> <http://example.com/u> .
```

I think currently this library does not support creating n-quads in a canonical form like that. It may also be the changes needed for support are not vast, but I am not sure of that either. Maybe this issue could be used to enumerate changes needed to create canonical RDF Datasets by way of using this library (where make modifications, how and so on).

Currently there are projects on-going in EU and elsewhere where projects and standards implementations seem to be choosing between canonical JSON or canonical RDF when using self-sovereign identity concepts: decentralized identifiers, verifiable credentials, verifiable presentations and so on. Since signing documents is a core technical use case, inteoperability with .NET tools would be a large benefit.

Likely the canonical example implementation is at https://github.com/digitalbazaar/rdf-canonize.

kal:
The main work here would be to implement the canonicalization algorithms in section 4 of the specification. The serialization of the canonicalized quads should be relatively easy to implement, but the canonicalization steps will take a bit more work and testing.

There is also a [test suite](https://w3c.github.io/rdf-canon/tests/) that should be integrated into the tests for the implementation.

zotanmew:
[...]

PR #618, opened 2024-01-19
Add RDF Canonicalization support

zotanmew:
This PR adds support for [RDF canonicalization](w3c-ccg.github.io/rdf-dataset-canonicalization/spec/), as well as the respective [test suite](https://w3c.github.io/rdf-canon/tests/).

Tasks remaining:
- [x] Fix test 60c (unicode/encoding)
- [x] Cleanup code
- [x] Integrate with JSON-LD handlers
- [x] Fix handling of JSON-LD datetime objects
- [x] Fix handling of URI nodes with a trailing slash in the path that also have an anchor

If I should refactor this to make it more in like with the rest of the codebase, please advise me on which patterns I've missed.


zotanmew:
Thanks to @deviant’s help, all 84 tests are now passing, and a bunch of bugs in the existing codebase were fixed as well :)

Libraries/dotNetRdf.Core/Writing/Formatting/NTriplesFormatter.cs
```diff
@@ -271,7 +277,19 @@ public override string FormatUri(Uri u)
...
 
-            return FormatUri(Rfc3987Formatter.EscapeUriString(u.ToString()));
+            var uriString = u.ToString();
+            if (uriString.EndsWith("/") && !u.OriginalString.EndsWith("/"))
+                uriString = uriString.Substring(0, uriString.Length - 1);
+            else if (!uriString.EndsWith("/") && u.OriginalString.EndsWith("/"))
+                uriString += '/';
+            else if (uriString.Contains("/#") && !u.OriginalString.Contains("/#"))
+                uriString = uriString.Replace("/#", "#");
+            else if (uriString.Contains("/#") && !u.OriginalString.Contains("/#"))
```
kal, line 287: This looks like a duplicate of lines 285-286

kal (changes requested):
Hi @zotanmew and @deviant. Apologies for the delay in getting this PR reviewed! Thanks very much for this contribution. It all looks great and I really appreciate that you added the full test suite too. I just have one minor change request. I'm going to approve the CI run too now

Libraries/dotNetRdf.Core/Writing/Formatting/NTriplesFormatter.cs
zotanmew: Oh, oops, pasted that block twice ^^;

Should be fixed.

zotanmew:
I just realized that the official version of the specification is hosted at https://www.w3.org/TR/rdf-canon/#dfn-rdfc-1-0, while I implemented it based on https://w3c-ccg.github.io/rdf-dataset-canonicalization/spec/. The algorithm they specify is the same, but the former is much more readable. I'm gonna have another look to make sure this is compliant with the actual spec, despite the tests passing.

Also, any opinions on the exposed public methods for this? Are they good as-is? Is it fine that this requires passing an already expanded document? (Seemed like the best thing to do given software might want to process the expanded document on top of getting it canonicalized, to not have to do the work twice)

zotanmew:
I've been referred to https://www.w3.org/TR/rdf-canon/#changes-from-fpwd, which does have changes (albeit minor ones) from the version I implemented. I shall get to work.

kal:
@zotanmew I think it is fine to have the API require an expanded document as input. Convenience methods could always be added later if there is a need. There are a couple of fixes needed to the formatter classes to avoid regressions. I've got the fixes locally but don't have permission to push them to your fork. I'll attach them as a unified diff file here.
[formatter_fixes.patch.gz](https://github.com/dotnetrdf/dotnetrdf/files/13998514/formatter_fixes.patch.gz)

zotanmew:
Thank you so much! I'll finish up the correct algo implementation (now with comments, revised first degree quads pass & compliant return types), merge your patch and then push all of that :)

zotanmew:
All done :)

kal (approved):
All looks great to me, thank you @zotanmew and @deviant !

zotanmew:
I think the workflow isn't loading the new tests, did I forget something or is this in the workflow definition? (regardless, they all pass on my machine)

zotanmew:
Hmm, one of them crashed o.O (https://github.com/dotnetrdf/dotnetrdf/actions/runs/7595621014/job/20688550712)

kal:
Its in the workflow definition in build.yaml. There's a list of the test projects to run for both the `test` job and the `test-linux` job. Ca you add the test suite project there?

kal:
> Hmm, one of them crashed o.O (https://github.com/dotnetrdf/dotnetrdf/actions/runs/7595621014/job/20688550712)

Yeah there are intermittent errors with the full text query tests. I *think* it may be a race condition with some of the other tests but haven't had time to get to the bottom of it. I've triggered a re-run as that can usually clear it.

zotanmew:
Hm, same test again :/
Since it's crashing it's unlikely to be a regression caused by this, right?

kal:
Yep same build issue as before. I've created a ticket to remind myself to look into this at some point because its really annoying :laughing: 

All checks are passed, so I'll merge now.
Awesome work, thank you very much! :heart:

zotanmew:
Thank you so much! This should probably close #615 as well :)

kal:
> Thank you so much! This should probably close #615 as well :)

Definitely/! I'll write a little bit of documentation for the feature tomorrow and then close the ticket once that is done. :)

PR #621, opened 2024-01-20
DeepClone input JArray on JSON-LD Canonicalize

zotanmew:
If we don't do this, the JsonLdParser.Load() method will mess with the original object, which is likely not desirable.

(I just noticed this during testing, apologies for missing it in the initial PR)


kal (approved):

kal:
Good spot, thanks @zotanmew !

PR #625, opened 2024-01-23
Basic Quads API

kal:
* Add `Quad` type to represent a statement in a graph.
* Add  `IEnumerable<Quad> Quads` property to `ITripleStore` to iterate all quads in the loaded graphs of the store.
* Add `Assert(Quad)` and `Retract(Quad)` to `ITripleStore` to add / remove quads in the store.
* Add `GetQuads()` to `ITripleStore` to get a filtered enumeration of quads in the loaded graphs of the store.

There is a basic implementation of all of these properties/methods in `BaseTripleStore`

### DIFF
--- a/Libraries/dotNetRdf.Core/Core/BaseTripleStore.cs
+++ b/Libraries/dotNetRdf.Core/Core/BaseTripleStore.cs
@@ -121,7 +121,39 @@ namespace VDS.RDF
         public abstract IUriFactory UriFactory { get; }
         #endregion
 
+        #region Assert/Retract Quads
+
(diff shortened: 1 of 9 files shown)


kal:
@langsamu - as you made a start on this I wonder if you could take a look at this PR. Its slightly different from the work you have in progress on your draft PR but I think fulfils the basic needs. I was in two minds about whether to model the Quad as Triple+IRefNode or as four nodes and in the end went with the latter, but could go back to the former as the internal implementation. I think its a finely balanced decision either way though.

PR #626, opened 2024-02-08
Fix querying and constructing langStrings

ddeschepper:
Edited the `ProcessLangString` method of the `BaseExpressionProcessor` to return a StringNode with language if it's dealing with a node with datatype `xsd:string`. This used to throw an `RdfQueryException`, because a language cannot be specified in combination with a typed literal. The exception here is of course the type `xsd:string`, that is implied when a string literal has no explicit type.

Additionally, `GetNode()` int the `ConstructContext` needs to handle the langString case before the other datatypes, because `.DataType` is not null in case of  a langStgring.


kal:
That's great, thank you @ddeschepper! 🎉

PR #633, opened 2024-04-22
 Turtle / TriG QNames' localname may have leading digits

### LINKED ISSUE #632: QName with leading digit in local part rejected
When parsing a trig file that includes a qname with leading digit in the local part I get this error message:
`Unexpected input 'exdata:1test', expected a QName, the 'a' Keyword or a Plain Literal`

To reproduce, run
```
var trigParser = new TriGParser();
var tStore = new TripleStore();
trigParser.Load(tStore, new StringReader("@prefix exdata: <https://example.com/data/>. exdata:test1 { exdata:1test a <http://example.com/Thing> .}"));
```

The standard at https://www.w3.org/TR/turtle/ allows this: 
```
 Prefixed names are a superset of XML QNames. They differ in that the local part of prefixed names may include:

    leading digits, e.g. leg:3032571 or isbn13:9780136019701
```

markus-ap:
According to W3C Trutle / TriG may have leading digits in their local name.
https://www.w3.org/TR/turtle/#h_note_2

I added support to allow for digits as starting characters for Turtle / TriG.

In addition I noted that there is a check that disallows more than one colon in a QName. This should be allowed per the above documentation from W3C, but I did not undertake this fix in this PR.

This fixes #632.

I used the test from the issue above as the base for a unit test.

### DIFF
--- a/Libraries/dotNetRdf.Core/Parsing/TurtleSpecsHelper.cs
+++ b/Libraries/dotNetRdf.Core/Parsing/TurtleSpecsHelper.cs
@@ -24,6 +24,7 @@
 // </copyright>
 */
 
+using AngleSharp.Text;
 using System;
 using System.Text.RegularExpressions;
 using VDS.RDF.Parsing.Tokens;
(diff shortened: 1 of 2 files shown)


kal:
Thanks @markus-ap

kal:
Added #634 to track the additional issue that you reported.

PR #671, opened 2024-12-18
Add dataGraph to implementations of INode.ObjectsOf(INode, IGraph).

aemnathanclinton:


kal:
Thank you @aemnathanclinton ! I've add a repro as well for the issue you reported. The full PR is now on #672

PR #674, opened 2024-12-26
Remove redundant test

langsamu:
Removing a test (related to RDF list functionality) that is exactly the same as the one before it.


kal:
Thanks @langsamu !

PR #677, opened 2025-01-14
The possible matches should be SetIDs from the Right Hand Side

joost-morsink:
I think intersecting or setting the SetIDs from the left hand side to the possible matches is a mistake, as the left join should consider SetIDs from the right hand side (other).

### DIFF
--- a/Libraries/dotNetRdf.Core/Query/Algebra/BaseMultiset.cs
+++ b/Libraries/dotNetRdf.Core/Query/Algebra/BaseMultiset.cs
@@ -362,7 +362,7 @@ namespace VDS.RDF.Query.Algebra
                 else
                 {
                     // Don't forget that a null will be potentially compatible with everything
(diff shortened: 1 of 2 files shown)


joost-morsink:
So... the failing test should not even hit the modified code, can we rerun the tests?

kal:
Yeah, that looks correct to me from eyeballing the code., but I'd feel safer merging if there was a test that failed before the change and passes with the change implemented. Do you have a simple test case for this issue?

joost-morsink:
Actually I do have quite a complicated test, which does way too much. I know exactly what the problem is so it won't be much trouble to implement a small test. It might take a few days though. I'll rebase the whole thing to a commit with failing test and a commit with succeeding test.

kal:
That would be great, thanks @joost-morsink !

kal:
That's perfect, thank you @joost-morsink !

Issue #675, opened 2025-01-07
Add "Safe Mode" option


BigBlueHat:
The JSON-LD Community Group and Working Group at the W3C have begun discussing standardizing the "safe mode" option that is in `jsonld.js` and many other JSON-LD libraries: https://github.com/digitalbazaar/jsonld.js?tab=readme-ov-file#safe-mode

It's an essential feature when creating documents where developers need to stop the train if a document is not fully described by it's context--as is the case for secure document formats like [Verifiable Credentials](https://www.w3.org/TR/vc-data-model-2.0/) or anything signed with [Data Integrity](https://www.w3.org/TR/vc-data-integrity/).

This would be a huge help to downstream VC implementers depending on this library.

Thanks for considering it!
🎩

kal:
@BigBlueHat - do you have a pointer to any early specification of what safe mode is? A quick Google turned up a couple of mentions in the Community Group meeting minutes, but no real detail.

kal:
I've documented the SafeMode support implemented by @sdml . I think this is enough to close this ticket for now. Happy to take further tickets if anyone spots missing/unexpected behaviours.

PR #685, opened 2025-01-28
SafeMode for JsonLd processing

sdml:
In relation to issue #675, I have created a "workaround" that could serve as a starting point for developing a proper safe mode feature for JSON-LD processing.


kal:
That's great, thank you @sdml !

kal:
I've tried to address the test failures locally, and have had to move this over to a new PR with those changes: #687

PR #689, opened 2025-03-09
Update W3C JSON-LD 1.1 test suite

### LINKED ISSUE #688: W3C json-ld test suite: expansion t0131 fails
I've been trying to build my own implementation (not in C#) of the JSON-LD 1.1 spec. Since I've struggled with some parts of the spec, I've used the dotNetRdf one to help me figure some things out since it seems to be the only implementation that implements the recommended/final version instead of previous drafts. So first of all, thank you very much!

However, both myself and dotNetRdf seem to fail on t0131 in the official expansion test suite: https://w3c.github.io/json-ld-api/tests/expand-manifest.html. It's defined as `specVersion: 1.1` and does not specify a `processingMode`, so the result should be the same in both 1.0 and 1.1 processing.

The input is:

```json
{
  "@context": {
    "@version": 1.1,
    "@base": "https://example.org/",
    "@vocab": "https://example.net/ns#",
    "rdf": "http://www.w3.org/1999/02/22-rdf-syntax-ns#",
    "statement": {"@reverse": "rdf:subject", "@container": "@index", "@index": "predicate"},
    "predicate": {"@id": "rdf:predicate", "@type": "@vocab"},
    "term": {"@id": "rdf:object", "@type": "@vocab"},
    "addedIn": {"@type": "@id"}
  },
  "@id": "item/1",
  "statement": {
    "rdf:type": {"term": "A", "addedIn": "v1"}
  }
}
```

The expected output is:

```json
[
  {
    "@id": "https://example.org/item/1",
    "@reverse": {
      "http://www.w3.org/1999/02/22-rdf-syntax-ns#subject": [
        {
          "https://example.net/ns#addedIn": [
            {
              "@id": "https://example.org/v1"
            }
          ],
          "http://www.w3.org/1999/02/22-rdf-syntax-ns#object": [
            {
              "@id": "https://example.net/ns#A"
            }
          ],
          "http://www.w3.org/1999/02/22-rdf-syntax-ns#predicate": [
            {
              "@id": "http://www.w3.org/1999/02/22-rdf-syntax-ns#type"
            }
          ]
        }
      ]
    }
  }
]
```

But dotNetRdf gives me:

```json
[
  {
    "@id": "https://example.org/item/1",
    "@reverse": {
      "http://www.w3.org/1999/02/22-rdf-syntax-ns#subject": [
        {
          "http://www.w3.org/1999/02/22-rdf-syntax-ns#object": [
            {
              "@id": "https://example.net/ns#A"
            }
          ],
          "https://example.net/ns#addedIn": [
            {
              "@id": "https://example.org/v1"
            }
          ],
          "@index": "rdf:type"
        }
      ]
    }
  }
]
```

That's based on this little program:

```csharp
using Newtonsoft.Json.Linq;
using VDS.RDF.JsonLd;

namespace Sharpnet {
    class Program
    {
        static void Main(string[] args)
        {
            JObject input = JObject.Parse(File.ReadAllText("expand/0131-in.jsonld"));
            JArray result;
            result = JsonLdProcessor.Expand(input, new());
            Console.WriteLine(result);
        }
    }
}
```

(I apologise for any C# crimes, it's the first time I've used the language).

I believe the issue stems from the fact that during Term Creation, the `@index` field on `statement` is never read, because the spec has us handle `@reverse` first and then tells us to return (https://www.w3.org/TR/json-ld11-api/#create-term-definition step 13.7). I've checked the errata on the 1.1 spec, but I can't find anything related to it.

However, the JSON-LD playground does produce the expected output, and I suppose draft and final release of the spec ought to handle this case in the same way.

zotanmew:
This PR updates the W3C JSON-LD 1.1 test suite to the latest version as per [w3c/json-ld-api](https://github.com/w3c/json-ld-api/).

It also fixes all the issues that caused the new tests to fail.

Resolves #688


zotanmew:
I absolutely do not understand test toRdf/li12, it has an invalid `@base` property but expects the implementer not to throw, but to instead set it to null, in direct contradiction with context processing step 5.7.5, which states

> an invalid base IRI error has been detected and processing is aborted.

zotanmew:
The commits I just pushed pass all but 3 tests. One for the reason mentioned in the previous comment, One due to `System.InvalidOperationException: Cannot access child value on Newtonsoft.Json.Linq.JValue.`, and one due to an output mismatch. I'll keep digging.

zotanmew:
Fixed one more bug, getting the broken test count down to 2. I'm honestly not sure how to fix the output mismatch (compact test t0114), and am fairly sure the toRdf test tli12 is just broken.

zotanmew:
The broken test is known and was reported years ago - https://github.com/w3c/json-ld-api/issues/533

zotanmew:
Figured it out. This should pass all tests now.

kal (approved):
All looks great. Thanks for your work on this @zotanmew !

PR #694, opened 2025-04-16
Proper `∀` for `$PATH` substitution in `sh:select`

### LINKED ISSUE #692: SHACL: Improper bindings for SPARQL-based Constraint Components with sh:propertyValidator
### Issue
When using a SPARQL-based Constraint with a sh:propertyValidator, the SPARQL query executed is not the desired one.
All predicates are replaced with the given $PATH even when the Predicate.Variables list is empty, which should not be the case.

### SHACL Example
```
@prefix ex: <http://example.org/> .
@prefix foaf: <http://xmlns.com/foaf/0.1/> .

ex:knowsNobodyShape
    sh:targetClass foaf:Person ;
    sh:property [
        sh:path foaf:knows ;
        ex:param 42 ;
    ] .
    
ex:knowsNobodyComponent
    a sh:ConstraintComponent ;
    sh:parameter [
        sh:path ex:param ;
    ] ;
    sh:propertyValidator [
        sh:select """
            PREFIX foaf: <http://xmlns.com/foaf/0.1/>
            SELECT DISTINCT $this
            WHERE {
                $this a foaf:Person.
                $this $PATH ?someone .
            }
        """ ;
    ] .
```
Generated Query After Binding
```
PREFIX ex: <http://example.org/>
PREFIX foaf: <http://xmlns.com/foaf/0.1/>
SELECT DISTINCT ?this WHERE
{ 
  BIND(42 AS ?param) . 
  BIND(ex:shapeGraphPropertySELECT AS ?shapesGraph) . 
  BIND(_:autos3 AS ?currentShape) . 
  BIND(ex:Alice AS ?this) . 
  ?this foaf:knows foaf:Person . 
  ?this foaf:knows ?someone . 
}
```
🔍 Location of Bug
Select.cs

Line: 134

Problematic Code:

csharp
Copier
Modifier
if (pattern.TriplePatterns[i] is TriplePattern triplePattern && 
    triplePattern.Predicate.Variables.All(x => x.Equals("PATH")))
🐞 Bug Explanation
The condition above evaluates to true even when triplePattern.Predicate.Variables is empty.
This happens because .All() on an empty collection returns true, which is logically incorrect for this context.

✅ Proposed Fix
Update the condition to check if the Predicate.Variables list is not empty before applying .All().
Fixed Code:

csharp
Copier
Modifier
if (pattern.TriplePatterns[i] is TriplePattern triplePattern && 
    triplePattern.Predicate.Variables.Any() &&  
    triplePattern.Predicate.Variables.All(x => x.Equals("PATH")))

langsamu:
# What

Fixes #692, which identified a mistake in the implementation of `SELECT`-based validators in `SPARQL`-based Constraint Components.

# Why

The SHACL spec says this in its usual poetic brevity:
> [...] prior to execution substitute the variable PATH where it appears in the predicate position of a triple pattern with a valid SPARQL surface syntax string of the SHACL property path specified via sh:path at the property shape. [...]

-- [6.3 Validation with SPARQL-based Constraint Components](https://www.w3.org/TR/shacl/#constraint-components-validation)

We need `$PATH` in `sh:select` so we can refer to the predicate our constraint was anchored on.

# How

This change [fixes](https://github.com/dotnetrdf/dotnetrdf/pull/694/files#diff-ce49e2948c178ea4525a643a75d67742b7e43e391edc45a54a8b97d928cacbcfL134-R139) the implementation of the algorithm specified in the quote above.
The mistake (thx @Ikeragnell) is a well-known one about universal quantification over empty sets, summarized nicely in [my favourite quote from the W3C website](https://www.w3.org/TR/owl2-primer/#Property_Restrictions:~:text=There%20is%20one%20particular%20misconception%20concerning%20the%20universal%20role%20restriction).

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -5,6 +5,7 @@ Next release
 ------------
 
 NEW: Added a new `SafeMode` option for the JSONLD Parser. Setting this option to true will cause additional warnings to be raised when an invalid type, subject, predicate or object is encountered when parsing JSON-LD rather than silently ignoring them. Thanks to @BigBlueHat for the suggestion and @sdml for the initial implementation (#675)
+FIX: Fix for `$PATH` substitution in `SELECT`-based validators in `SPARQL`-based Constraint Components. Thanks to @Ikeragnell for the report and @langsamu for the fix. (#694)
  
 3.3.2
 -----
--- a/Libraries/dotNetRdf.Shacl/Constraints/Select.cs
+++ b/Libraries/dotNetRdf.Shacl/Constraints/Select.cs
@@ -131,11 +131,12 @@ namespace VDS.RDF.Shacl.Constraints
         {
             for (var i = 0; i < pattern.TriplePatterns.Count(); i++)
             {
-                if (pattern.TriplePatterns[i] is TriplePattern triplePattern && triplePattern.Predicate.Variables.All(x=>x.Equals("PATH")))
-                {
-                    pattern.TriplePatterns.RemoveAt(i);
-                    pattern.TriplePatterns.Insert(i, new PropertyPathPattern(triplePattern.Subject, path, triplePattern.Object));
-                }
+                if (pattern.TriplePatterns[i] is not TriplePattern triplePattern) { continue; }
+                if (triplePattern.Predicate is not VariablePattern variablePattern) { continue; }
+                if (variablePattern.VariableName != "PATH") { continue; }
+
+                pattern.TriplePatterns.RemoveAt(i);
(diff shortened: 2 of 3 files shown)


langsamu:
New test [passes](https://github.com/dotnetrdf/dotnetrdf/actions/runs/14496710036/job/40666192009?pr=694#step:3:55) after fix:
image

langsamu:
[Failure](https://github.com/dotnetrdf/dotnetrdf/actions/runs/14616067940/job/41004486509?pr=694#step:3:165) is unrelated.

BTW it seems wrong to load an external resource in tests:

image

kal:
@langsamu - if you merge the changes on main then hopefully the one failing test will be fixed.

kal (approved):

PR #713, opened 2025-06-04
File-scoped namespaces

langsamu:


kal (approved):
Took a while to go through this but all looks good to me.

PR #714, opened 2025-06-10
Add Dependabot

hellikopter:
Use Dependabot to [automate dependency version updates](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuring-dependabot-version-updates).


kal (approved):
This could be really helpful, thanks!

PR #785, opened 2025-11-15
Deprecate global extensions

langsamu:
# Summary

This change deprecates a few public global extenion methods that add convenience functionality to `object` under our root namespace:

```cs
IEnumerable<T> AsEnumerable<T>(this T item)
bool IsDisjoint<T>(this IEnumerable<T> x, IEnumerable<T> y)
IEnumerable<T[]> ChunkBy<T>(this IEnumerable<T> source, int size)
string ToSafeString(this object obj)
string ToSafeString(this Uri u)
string ToString(this Uri u, IUriFormatter formatter)
```

# Description

## Why

I have several problems with these extensions:
- They are too widely scoped in terms of the types they extend, like `object`. This makes them apply to everything. Always.
- They are too widely scoped in terms of their namespace, `VDS.RDF`. This is our root namespace, containing fundamentals like `IGraph` and `INode`. It is almost certain to be in scope when using any part of dotNetRDF.
- They are too widely scoped in terms of accessibility. They are a burden to maintain, test, document and version.
- They are irrelevant to functionality of this library.
- They implement patterns that seem convenient but are not necessarily robust. Anything that takes away the pain of dealing with nulls ends up causing even more pain.
- They implement patterns that are better served by modern language features.

I do see value in consistent application of patterns across our codebase. But for these examples, I think the cost is greater than the benefit.

## What

The change I made is to
- inline the functionality of the extensions to every call site and
- deprecate (warning) the methods,
- offering workarounds in documentation comments.

# History

- Some of the methods used to be `internal` but made public in 9369469, probably due to package restructure.
- Another goes about as far back as possible to a35ad48.

# Elsewhere

There are some interesting ideas in [Framework Design Guidelines](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/) that are relevant to extension methods and my objections to our usage of them,

- image
- image
- image
- image

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -1,6 +1,11 @@
 ﻿Change Log
 ==========
 
+3.4.2
+-----
+
+FIX: Deprecated several globally scoped extension methods. They are now marked obsolete usable.
+
 3.4.1
 -----
 


Libraries/dotNetRdf.Core/Core/SubGraphMatcher.cs
```diff
@@ -530,7 +530,7 @@ private bool TryBruteForceMapping(IGraph subgraph, IGraph parent, Dictionary<INo
         // Populate existing Mappings
         foreach (KeyValuePair<INode,INode> fixedMapping in _mapping) 
         {
-            possibleMappings.Add(fixedMapping.Key, new List<INode>(fixedMapping.Value.AsEnumerable<INode>()));
+            possibleMappings.Add(fixedMapping.Key, [fixedMapping.Value]);
```
langsamu: Collection expressions are smart enough to go straight to lists.

Libraries/dotNetRdf.Core/Query/Algebra/Extend.cs
```diff
@@ -95,14 +95,14 @@ public IEnumerable<string> Variables
     {
         get 
         {
-            return InnerAlgebra.Variables.Concat(VariableName.AsEnumerable()); 
+            return [.. InnerAlgebra.Variables, VariableName];
```
langsamu: Using a [spread element](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/collection-expressions#spread-element) to simplify concatenation.

Libraries/dotNetRdf.Core/Query/Algebra/ParallelJoin.cs
```diff
@@ -43,7 +43,7 @@ public class ParallelJoin : IJoin
     /// <param name="rhs">Right Hand Side.</param>
     public ParallelJoin(ISparqlAlgebra lhs, ISparqlAlgebra rhs)
     {
-        if (!lhs.Variables.IsDisjoint(rhs.Variables)) throw new RdfQueryException("Cannot create a ParallelJoin between two algebra operators which are not distinct");
+        if (lhs.Variables.Intersect(rhs.Variables).Any()) throw new RdfQueryException("Cannot create a ParallelJoin between two algebra operators which are not distinct");
```
langsamu: I opted for this pattern instead of the one it replaces. To my ears,
> intersection is empty

is more intuitively understandable than
> all of them are not contained

```cs
return x.All(item => !y.Contains(item));
```

Libraries/dotNetRdf.Core/Query/Ordering/OrderByVariable.cs
```diff
@@ -102,8 +102,8 @@ public override IEnumerable<string> Variables
         get
         {
             return Child != null
-                ? Variable.AsEnumerable().Concat(Child.Variables).Distinct()
-                : Variable.AsEnumerable();
+                ? ((IList<string>)[Variable, .. Child.Variables]).Distinct()
+                : [Variable];
```
langsamu: Not the nicest syntax when the target type for the collection expression is unknown.

Libraries/dotNetRdf.Core/Query/LeviathanQueryProcessor.cs
```diff
@@ -2216,7 +2216,7 @@ private IEnumerable<SparqlQuery> GetRemoteQueries(SparqlEvaluationContext contex
         else
         {
             // Split bindings in chunks and inject them
-            foreach (ISet[] chunk in bindings.ChunkBy(100))
+            foreach (ISet[] chunk in ChunkBy(bindings, 100))
```
langsamu: That extension method had this single call site and it was not something I could replace with a single line, so I inlined it as a utility method into this file instead.

Libraries/dotNetRdf.Core/Storage/StorageHelper.cs
```diff
@@ -174,7 +174,7 @@ private static string GetStatusLine(WebException webEx)
             {
                 return "(HTTP " + (int)httpResponse.StatusCode + " " + httpResponse.StatusDescription + ")";
             }
-            return webEx.Status.ToSafeString();
+            return webEx.Status.ToString();
```
langsamu: `Status` is an `Enum`, so it's never `null`.

Libraries/dotNetRdf.Core/Update/Commands/InsertCommand.cs
```diff
@@ -85,7 +85,7 @@ public override bool AffectsSingleGraph
             var affectedUris = new List<string>();
             if (TargetGraph != null)
             {
-                affectedUris.Add(TargetGraph.ToSafeString());
+                affectedUris.Add(TargetGraph.ToString());
```
langsamu: Guarded [above](https://github.com/dotnetrdf/dotnetrdf/pull/785/files#diff-30ec8668835c12b30b940360fa33deed21f59b63ca2dd9adca71b48e9c74af9dR86).

langsamu (commented):
This review contains explanations for the first occurrance of each pattern of change.

I recommend reading the deprecation documentation comments on the [extension methods](https://github.com/dotnetrdf/dotnetrdf/pull/785/files#diff-f5eb90d725255fc47ab9b387403a090c62ee0124294e03b05e68d8f5b637277f) first.

Libraries/dotNetRdf.Core/Query/Algebra/ISet.cs
```diff
@@ -281,7 +281,7 @@ public override string ToString()
         var count = 0;
         foreach (var var in Variables.OrderBy(v => v))
         {
-            output.Append("?" + var + " = " + this[var].ToSafeString());
+            output.Append("?" + var + $" = {this[var]}");
```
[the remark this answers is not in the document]
kal: I think that could safely be done in a separate pass through.

Issue #845, opened 2026-02-04
Memory leak due to not unsubscribed event handlers


bruluk5w:
### How to reproduce
When you use the InMemoryExtensions to execute a SPARQL query on a graph, one can observe that the retained memory after each query linearly grows. The following is a minimal example, where each iteration through the loop increases the total memory (12 bytes on my system). Note that neither the actual query nor the contents of the graph matter.

```C#
using VDS.RDF;
IGraph graph = new Graph();
for (var i = 0; i < 1000; i++) {
    GC.Collect();
    GC.WaitForPendingFinalizers();
    Console.WriteLine(GC.GetTotalMemory(true));
    if(graph.ExecuteQuery("CONSTRUCT {?s <http://example.org/somepredicate> ?o } WHERE { ?s <http://example.org/predicate/somepredicate> ?o }") is IDisposable d) {
        d.Dispose();
    }
}

graph.Dispose();
return 0;
```
Below is a screenshot of a comparison between a memory snapshot before and after a query execution:

<img width="1412" height="347" alt="Image" src="https://github.com/user-attachments/assets/48416b0d-7da6-4405-8204-2e25f2466bdc" />

When looking at the top reference paths that lead to the retained instances of `TripleStore` one can see that the objects are still kept alive through event listeners that have not been unregistered:

<img width="416" height="541" alt="Image" src="https://github.com/user-attachments/assets/d7bd7d4e-4005-4f02-8b50-1e94b326a65b" />

While it does not consume a large amount of memory, it is permanent and also increases the mutation latency for retract and assert operations on the graph. The graphic below shows the situation in a scenario when one tries to retract and assert triples while querying in parallel via `ExecuteQuery`. Note that the large spike in the image below is only a side-effect caused by temporary high load on my machine.
 
<img width="983" height="395" alt="Image" src="https://github.com/user-attachments/assets/8fe8225d-8757-4676-b434-bb84b58b538b" />
The line labeled "Other" in the image above is low when no SPARQL queries are issued (waiting time on the lock in my code is none). And it is high when SPARQL queries are issued (there is waiting time on the lock until the SPAQL query finished). We can observe that the latency for retract and assert does not grow when no SPARQL queries are issued and it grows when SPARQL queries are issued.

Below there are two flamegraphs. The first shows the normalized timeline of queries after starting the program. The second shows the normalized timeline after letting the program run for an extended amount of time:

[...]

PR #846, opened 2026-02-04
Fix/deregistration of event handlers and proper disposal

bruluk5w:
These changes fix the problem documented in issue #845. 

These changes do not enforce the proper usage of `IDisposable.Dispose` but allow the execution of SPARQL queries without causing the memory leak described in #845.


atodkar:
@kal @rvesse is it possible to merge this change and prepare a patch release? How do we request a patch with this fix?

rvesse:
> @kal @rvesse is it possible to merge this change and prepare a patch release? How do we request a patch with this fix?

I haven't been involved in the project, nor touched any .Net code in ~10 years now, so that's really a question for @kal and the other maintainers

kal:
Thanks for the report and the fix. I've reviewed the code and it looks good to my eyeballs. I've changed the PR target to a new maintenance branch for the 3.x codebase as `main` is now for 4.0 development work. I'll let the tests run and then merge to the maintenance branch to make a 3.5.1 release. I'll also port the changes over into `main` so the fixes are applied to 4.0 work.

kal:
Ah, you already had the latest code. I'll try doing it the other way around. Test against main and if all is OK port the changes back over to the maintenance branch.

Libraries/dotNetRdf.Core/Query/LeviathanQueryProcessor.cs
```diff
@@ -4547,4 +4547,22 @@ private static IEnumerable<T[]> ChunkBy<T>(IEnumerable<T> source, int size)
             buffer.Clear();
         }
     }
+
+    protected virtual void Dispose(bool disposing)
+    {
+        if (disposing)
+        {
+            _lock?.Dispose();
+            if (_dataset is IDisposable d)
```
kal: This doesn't seem right. The LeviathanQueryProcessor doesn't own the dataset and should not be disposing it here.

kal (changes requested):

Libraries/dotNetRdf.Core/Query/LeviathanQueryProcessor.cs
bruluk5w: Yes, you are right, I added the `using` statements already in the `InMemoryExtensions` to dispose the dataset there. I will remove the line here that disposes the dataset inside the `LeviathanQueryProcesssor`.

Libraries/dotNetRdf.Core/Query/Datasets/InMemoryDataset.cs
```diff
@@ -401,4 +401,19 @@ protected override void FlushInternal()
...
+
+    /// <summary>
+    /// Disposes The InMemoryDataset
+    /// </summary>
+    /// <param name="disposing">True if called via <see cref="Dispose"/>.</param>
+    protected override void Dispose(bool disposing)
+    {
+        if (disposing)
+        {
+            _store?.Dispose();
```
kal: I think this might also be an issue. The InMemoryDataset instances doesn't always own the store - some constructors allow it to be passed in. I think this class should only dispose the store if it created it. That could be determined by another private bool field that gets set only by constructors that create the store.

kal (changes requested):

Libraries/dotNetRdf.Core/Query/Datasets/InMemoryDataset.cs
bruluk5w: You are right, I'm committing  a fix.

Libraries/dotNetRdf.Core/Query/Datasets/InMemoryDataset.cs
```diff
@@ -39,43 +39,48 @@ public class InMemoryDataset
...
+        : this(g.AsTripleStore(), g.Name, disposeStore: true) { }
 
     /// <summary>
     /// Creates a new In-Memory dataset.
     /// </summary>
     /// <param name="store">In-Memory queryable store.</param>
-    public InMemoryDataset(IInMemoryQueryableStore store)
-        : this(store, false) { }
+    /// <param name="disposeStore">Whether the store should be disposed when the InMemoryDataset is disposed</param>
+    public InMemoryDataset(IInMemoryQueryableStore store, bool disposeStore = false)
```
kal: This change is what is causing the tests to fail. It is also a breaking API change because if you have an existing call `new InMemoryDataset(store, true)` that now goes to this constructor instead of the constructor below. If you remove `disposeStore` parameter and just always pass false for `disposeStore` that fixes the test.

kal (approved):
Thanks for working on this @bruluk5w ! Changes all look good now.

atodkar:
Thank you @kal for a quick turnaround on this issue. Would it be a lot if I request a patch on 3.4 relesae? 
We are not yet on the dotnet 10 and would be difficult for us to use this fix in a short while otherwise.
Appreciated your support on this.

kal:
Hey @atodkar, the 3.5.1 release targets.net standard 2.0, so it should be
fine to use on older versions of dotnet.

anandtodkar:
@kal thank you. Yes we tested and it worked.

PR #877, opened 2026-05-21
fix: handling of star and plus operators in SPARQL query

fennibay:
I see that star and plus operators are not working as expected, at least in SELECT queries..

This PR adds a test to demonstrate the problem, I don't have a solution yet.

For this knowledge graph:

```ttl
@prefix ex: <http://example.org/> .
ex:n1 ex:p1 ex:n2 .
ex:n2 ex:p1 ex:n3 .
ex:n3 ex:p1 ex:n4 .
```

And this query:

```sparql
PREFIX ex: <http://example.org/>
SELECT ?n WHERE {
  ?n ex:p1* ex:n4 .
}
```

I expect four results: ex:n1, ex:n2, ex:n3, ex:n4, but it is currently returning 2 results: ex:n3, ex:n4.

When I change the `*` operator to `+`, I expect it to return three results, but it is returning only 1: ex:n3.

I checked with <https://atomgraph.github.io/SPARQL-Playground/> as a reference, that is behaving as I expect.


fennibay:
I added a fix candidate yesterday. The added unit tests are passing. Please review.

kal:
Thanks for the fix @fennibay !

PR #889, opened 2026-08-16
Bump AngleSharp from 1.4.0 to 1.5.0

langsamu:
Resolves #888.

PR #890, opened 2026-08-16
Remove AngleSharp

### LINKED ISSUE #888: AngleSharp 1.4.0 vulnerability
referenced by dotnetrdf.Core

https://github.com/advisories/GHSA-pgww-w46g-26qg

Image

langsamu:
Resolves #888.

Was looking at #889 but found that there is only a single usage of this dependency:
https://github.com/dotnetrdf/dotnetrdf/blob/d14063ff4802483cd4488796d063df733878cb93/Libraries/dotNetRdf.Core/Parsing/TurtleSpecsHelper.cs#L858

That extension method was implemented in AngleSharp 1.4.0 like [this](https://github.com/AngleSharp/AngleSharp/blob/v1.4.0/src/AngleSharp/Text/CharExtensions.cs#L218):
```cs
[MethodImpl(MethodImplOptions.AggressiveInlining)]
public static Boolean IsDigit(this Char c) => c >= 0x30 && c <= 0x39;
```

Using the [built-in method on `char`](https://learn.microsoft.com/en-us/dotnet/api/system.char.isdigit?view=netstandard-2.0) has slightly different semantics: [More than `0-9` are considered digits](https://dotnetfiddle.net/1GITTQ), e.g.
```cs
Console.WriteLine(char.IsDigit('६')); // true
```

So we might choose, instead of this change, to just inline the code above from AngleSharp.
OTOH the same `char.IsDigit` is used elsewhere in the same helper file, so it might be alright.

The method being changed seems to be our implementation of [`PN_CHARS`](https://www.w3.org/TR/turtle/#sec-grammar:~:text=%5B166s%5D-,PN_CHARS,-%3A%3A%3D).

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -14,7 +14,6 @@
 		<PackageVersion Include="System.Interactive.Async" Version="7.0.1" />
 		<PackageVersion Include="System.Linq.Async" Version="7.0.1" />
 		<PackageVersion Include="System.Threading.Channels" Version="10.0.7" />
-		<PackageVersion Include="AngleSharp" Version="1.4.0" />
 		<PackageVersion Include="System.ComponentModel.TypeConverter" Version="4.3.0" />
 		<PackageVersion Include="System.Configuration.ConfigurationManager" Version="10.0.8" />
 		<PackageVersion Include="System.Globalization.Extensions" Version="4.3.0" />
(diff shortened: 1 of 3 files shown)


kal:
Good spot @langsamu ! It would definitely be good to get rid of this dependency - I hadn't realised that it was no longer used anywhere but here after we did the migration away from old versions of .NET Framework.

using `char.Digit()` worries me because it validates a wider range of characters than the relevant Turtle specs demand. I see that it is actually used in several places in this file to validate the range `0-9`. Although .NET Standard 2.1 includes `char.IsAsciiDigit()` which would do what we want, I'm not ready to leave behind .NET Standard 2.0 just for that, so I think it might be better to defined our own version of that original AngleSharp method, maybe as `IsAsciiDigit(c)` for clarity.

kal:
Closed in favour of #892

